PR #31, opened 2015-01-11
Weapons usage

ThaPwned:
Updated how you handle weapons to an easier way


satoon101:
A weapons.entity.WeaponEntity class is already planned and in the works.  We will leave this pull request open as a reminder and possibly use some ideas from it, as well.  Thank you.

satoon101:
Actually, if you would like to continue your work on this, I have a few guidelines.

First, instead of the new class being players.weapons.Weapon, it needs moved to weapons.entity.WeaponEntity.  This will move it inline with entities.entity.BaseEntity and players.entity.PlayerEntity.

Next, instead of storing the player, only pass the index on instantiation, just like BaseEntity and PlayerEntity.  The player (or owner) of a weapon can easily change from one moment to the next, so storing that information is not a good idea.  Instead, use the m_hOwnerEntity property (which should be added a the "owner" attribute in all games that support it.

For reference, the csgo and orangebox engines both have the owner attribute already added:
* [csgo](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/data/source-python/entities/managers/engines/csgo/CBaseEntity.ini#L22)
* [orangebox](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/data/source-python/entities/managers/engines/orangebox/CBaseEntity.ini#L72)

You can then remove all of your properties of the Weapon class, as those are already properties of BaseEntity, which WeaponEntity should be inheriting from.

I have some fixes I need to make on the PlayerEntity side of things when it comes to weapons, but I will wait until you have made many of these changes.  Please make sure to re-pull, as we have now merged the entities_changes/entities_update branches into master, and that could have an affect on your changes.

satoon101:
Closing due to using #39 instead.

PR #34, opened 2015-01-17
Fixed server crash when precaching on server startup

olegtsvetkov:
Server crashes if engines.precache.Model object is being created on server load.

For example, I have simple plugin:
```python
from engines.precache import Model

test_model = Model('models/player/t_leet.mdl')
```
which is loaded through autoexec.cfg. When I start the server it crashes with this error:
```
Host_Error: CVEngineServer::PrecacheModel: 'models/player/t_leet.mdl' overflow, too many models
```

I don't know if there is the same problem with Decal and Generic class, but this fix will work for them too :smiley: 

Tested on CS:S, Linux (custom build from Jan 15)/Windows (official Jan 15 build).

### DIFF
--- a/addons/source-python/packages/source-python/engines/precache.py
+++ b/addons/source-python/packages/source-python/engines/precache.py
@@ -19,12 +19,11 @@ from path import Path
 # Source.Python Imports
 #   Core
 from core import AutoUnload
+from core import global_vars
 #   Engines
 from engines.server import engine_server
 #   Events
 from events.manager import event_registry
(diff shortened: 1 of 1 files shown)


olegtsvetkov:
O_o 
I didn't know that global_vars.map_name is None on server startup. Yes, your way is better!

satoon101:
Could you please verify that using map_name does indeed fix the problem and update the pull request if it does?

olegtsvetkov:
Done

PR #36, opened 2015-01-18
Fixed crash caused by ServerCommand registration without Help text.

olegtsvetkov:
When I register ServerCommand without help text in arguments my Linux CS:S server crashes.
```python

# No Help text
@ServerCommand('test_me')
def test_me_command(command):
    print("It works!")
```
This one works fine:
```python
@ServerCommand('test_me', 'Help text!')
```
Same problem when using `server_command_manager.register_commands` directly.

The segfault was caused by passing `NULL` to `strdup`.

Windows doesn't have this problem and I haven't tested how this fix works on Windows server, but everything should be fine... I think... ~~Any way I don't have Visual Studio installed.~~

### DIFF
--- a/src/core/modules/commands/server_commands_wrap.cpp
+++ b/src/core/modules/commands/server_commands_wrap.cpp
@@ -128,7 +128,7 @@ CServerCommandManager* CServerCommandManager::CreateCommand(const char* szName,
 		// Unregister the old command
 		g_pCVar->UnregisterConCommand(pConCommand);
 	}
-	else
+	else if( szHelpText != NULL )
 	{
 		// Store the given help text
 		szHelpTextCopy = strdup(szHelpText);


Ayuto:
Thanks for reporting! I haven't verified that, but if it's true we would have the same issue in this line. https://github.com/olegtsvetkov/Source.Python/blob/fd87e8398fe7801b37ede2c082cc868b4b74290b/src/core/modules/commands/server_commands_wrap.cpp#L125

olegtsvetkov:
I wasn't able to reproduce segfault on that line of code.
* **Not registered before command:** the [command is being created only one time](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/commands/manager.py#L52) and your line isn't being called at all.
* **Server commands:** SP just overrides them (tested with `listmaps`).
* **SourceMod commands:** both callbacks is being called.

Ayuto:
SP doesn't override the command if you use the C++ API.
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/commands/commands_wrap_python.cpp#L243
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/commands/server_commands_wrap.cpp#L247

However, the Python API shouldn't override them. So, this is probably something we need to fix.

olegtsvetkov:
As you can see [here](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/commands/server_commands_wrap.cpp#L129) if the command exists it would be overwritten.

Ayuto:
Yes, the command is unregistered there, but that doesn't mean the command object is completely deleted. The server would just stop calling the callback. In this [line](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/commands/server_commands_wrap.cpp#L138) pConCommand is passed to the constructor of CServerCommandManager. This variable can be either NULL (if the command wasn't registered previously) or a command object. Going just a few lines down in the code you can see [here](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/commands/server_commands_wrap.cpp#L147) that m_pOldCommand is set to the passed value. And later, when the command has been entered in the server console, CServerCommandManager::Dispatch() will be called. This will call your callback and if there was a command registered previously, it will be called as well as you can see [here](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/commands/server_commands_wrap.cpp#L247). So, both callbacks will be called. I doubt SM is doing something else.

satoon101:
I just tested and this indeed does crash on CS:S, but not on CS:GO, which I find odd.


As far as calling the old callback, it certainly seems to work for my testing:
```python

@ServerCommand('changelevel', 'This is not the help text')
def test_function(command):
    print('It works!!!')
```
After loading this code and using the **changelevel** command, the **It works!!!** is printed to the server console and then the level is changed.  And, when using
```
cvarlist changelevel
```
the original help text is shown and not the given one in the script.

satoon101:
As for line 125 which Ayuto mentioned, there does seem to be the potential for that to crash if GetHelpText returns NULL.  I have tested a few built-in server commands that do not have help text to see if I could replicate that crash, and I have been unsuccessful in doing so.  I imagine that their help text is an empty string.

I did test further by not passing a string for help text and then calling get_help_text on the command's instance, and it was an empty string (tested with type() as well).  So, maybe that gets set to a string no matter what if no value is passed.

olegtsvetkov:
Oops, sorry for bad info. I thought that `listmaps` is server command, not SM... And it looks like when I was checking it SM was disabled.

Ayuto:
I would be happy if you would also fix the line I mentioned. Imagine there is another server plugin, which created a command with NULL as the help text. We would run into the same issue. If you do that I will merge your pull request with the master.

olegtsvetkov:
@Ayuto, are you sure that it's really needed? @satoon101 said that nulled help text is being transformed into an empty string automatically by server.

I've also tried to register several same commands without help text and with one disabled [check](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/commands/manager.py#L52) (so the C part should be called all the time) and there was no crash.

Ayuto:
No, I'm not fully sure, but this can be a potential crash. Checking for NULL doesn't really hurt there and might save us from recieving random crashes.

olegtsvetkov:
But why we need checking for NULL if this variable can't be NULL at all?

Also if I'm wrong (I'm sure that not) and this line will cause crash, then it won't be random, because crash can only happend when the command is being registered.

Ayuto:
Like I said, I'm not fully sure if GetHelpText() can't return NULL.So, I did my own testings and was able to prove it with IDA.
```
.text:007EF950 ; ConCommandBase::Create(char  const*, char  const*, int)
.text:007EF950 _ZN14ConCommandBase6CreateEPKcS1_i proc near
.text:007EF950                                         ; CODE XREF: ConVar::Create(char  const*,char  const*,int,char  const*,bool,float,bool,float,void (*)(IConVar *,char  const*,float))+F5j
.text:007EF950                                         ; ConCommandBase::ConCommandBase(char  const*,char  const*,int)+Dj ...
.text:007EF950
.text:007EF950 arg_0           = dword ptr  8
.text:007EF950 arg_4           = dword ptr  0Ch
.text:007EF950 arg_8           = dword ptr  10h
.text:007EF950 arg_C           = dword ptr  14h
.text:007EF950
.text:007EF950                 push    ebp
.text:007EF951                 mov     ebp, esp
.text:007EF953                 push    ebx
.text:007EF954                 mov     eax, [ebp+arg_0]
.text:007EF957                 mov     edx, [ebp+arg_8]
.text:007EF95A                 mov     ebx, [ebp+arg_4]
.text:007EF95D                 mov     ecx, [ebp+arg_C]
.text:007EF960                 mov     byte ptr [eax+8], 0
.text:007EF964                 test    edx, edx
.text:007EF966                 mov     [eax+0Ch], ebx
.text:007EF969                 mov     ebx, (offset asc_948229+1)
.text:007EF96E                 cmovz   edx, ebx
.text:007EF971                 mov     [eax+14h], ecx
.text:007EF974                 and     ecx, 1
.text:007EF977                 mov     [eax+10h], edx
```

The interesting part is this:
```
; Test help text for NULL
.text:007EF964                 test    edx, edx

; Set m_pszName to the passed name value
.text:007EF966                 mov     [eax+0Ch], ebx

; Set ebx to a NULL terminator
.text:007EF969                 mov     ebx, (offset asc_948229+1)

; Set edx to the NULL terminator if edx is NULL
.text:007EF96E                 cmovz   edx, ebx

; Set m_nFlags to the passed flag value
.text:007EF971                 mov     [eax+14h], ecx

; Test the flags for flag 1
.text:007EF974                 and     ecx, 1

; Finally, set m_pszHelpString to edx
.text:007EF977                 mov     [eax+10h], edx
```
So, yes. GetHelpText() never returns NULL unless it has been modified (or the member variable itself) by memory hacking after initialization. Though, I should mention that I took this code from CS:S. It might be different on other games, but I don't think so.

PR #38, opened 2015-01-20
Added various of features

ThaPwned:


satoon101:
If you would, please use separate pull requests for separate issues/updates.  Since the weapons updates have nothing to do with the menus updates, we need them to be separate to discuss them separately.  Thanks.

PR #40, opened 2015-01-28
Fixed typo in entities/specials.py

olegtsvetkov:
One "t" is missing :)

https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/basetypes/basetypes_wrap_python.cpp#L365


Ayuto:
Thank you!

PR #41, opened 2015-02-23
Added decorator class used to handle delays

ThaPwned:


invincibleqc:
Can you please provide an example of any situation this could be useful?

ThaPwned:
It's basically just something to make it easier on the plugin developers. You'd do this:
```python
@Delayed
def do_something(name):
    ...
do_something(5, 'name')
```
Instead of:
```python
def do_something(name):
    ...
tick_delays.delay(5, do_something, 'name')
```

invincibleqc:
I don't really see how this make it easier, to be honnest. This is wrapping and duplicating the original API without adding anything except a different syntax. I also think this is anti-pythonic since it adds a required argument to the decorated functions which is not "physically" declared in their prototype.

satoon101:
I agree 100% with @invincibleqc on this.  Also, you can utilize TickRepeat to achieve the same thing:
```python
@TickRepeat
def do_something(name):
    ...

do_something.args = ('name', )
do_something.start(5, 1)
```

Not that I recommend the above code, but it is a more Pythonic solution.

PR #43, opened 2015-03-04
Added **kwargs support for menu translations

MarkusMeskanen:
All menus now support keyword arguments when translating them. This
includes _MenuData classes (texts and options), descriptions, and
titles.

### DIFF
--- a/addons/source-python/packages/source-python/menus/base.py
+++ b/addons/source-python/packages/source-python/menus/base.py
@@ -228,13 +228,17 @@ class _MenuData(object):
     All data types should inherit from this class.
     """
 
-    def __init__(self, text):
+    def __init__(self, text, **kwargs):
         """Initialize the instance.
 
         @param <text>:
         The text that should be displayed.
+
+        @param <kwars>:
+        Keyword arguments passed to the _translate_text function.
         """
         self.text = text
+        self.kwargs = kwargs
 
     def _render(self, player_index, choice_index=None):
         """Render the data.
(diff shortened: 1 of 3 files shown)


Ayuto:
There are a couple of issues with this implementation. At first the name "kwargs" is very ambiguous. A better name would be "tokens". The description would also need to be changed. However, this isn't a real problem, but one problem I see here is that you can't use "**kwargs", because every method has keywords. That would result in unusable token names like "value", "selectable", etc. Of course, this could be fixed by just using a new keyword "tokens=None".

However, I don't think this is the right approach, because these tokens are just statically passed to the translator function. If you want to update the tokens you would need to do something like this:
```
menu.description_tokens.update(...)
```
But that's just unecessary, because you can just do this:
```
menu.description.tokens.update(...)
```
So, the real problem we have here, is that you can't get a TranslationStrings object with an initialized tokens dictionary. We could simply work around that by adding this method to the LangStrings class:
```
    def get_translation_strings(self, key, **tokens):
        strings = self[key]
        strings.update(tokens)
        return strings
```
To show both methods in contrast:
```
# Your method
PagedMenu(description=my_strings['test'], description_tokens=dict(token1='abc', token2='bla'))

# Better, imo
PagedMenu(description=my_strings.get_translation_strings('test', token1='abc', token2='bla'))
```

MarkusMeskanen:
Oh, right :) Well I wasn't actually expecting this exact code to get accepted, I just gave my shot and was hoping people would come up with something. Apparently your solution was nothing similar, but much better.

However, it would be cool to get some sort of translation keywords for menus. I currently use the kwargs=None and self.kwargs = kwargs or {} solution, waiting for something better to show up.

Ayuto:
For now you can just move this method somewhere in your code. You can then do something like this:
```
def get_translation_strings(self, key, **tokens):
    strings = self[key]
    strings.update(tokens)
    return strings
    
menu = PagedMenu(description=get_translation_strings(my_strings, 'test', token1='abc'))
```
But I would still like to hear a third or fourth opinion on this.

MarkusMeskanen:
Yeah I figured, I already did that :)

MarkusMeskanen:
Yeah I figured, I already did that :) After re-reading your first comment a few times, that is.

Ayuto:
Since there was no more feedback, I added the get_strings() method. https://github.com/Source-Python-Dev-Team/Source.Python/commit/9507fd60db2309d058849cff7e3133b971d791d9

Issue #53, opened 2015-05-18
menus.PagedMenu() idea of "back" option


high-voltage:
It would be really nice if we could bind the "Back"-button in the first page of a PagedMenu() manually in our scripts. Example script of how it could look in usage:

```
def some_menu(index):
    menu = SimpleMenu()
    menu.append(SimpleOption(1, "some other menu", 1)
    menu.select_callback = some_menu_callback
    menu.send(index)

def some_menu_callback(menu, index, option):
    if option.value == 1:
        some_other_menu(index)

def some_other_menu(index):
    # create a paged menu with 2 pages
    menu = PagedMenu()
    for x in range(1, 10):
        menu.append(PagedOption(x, x))

    # this would tell the PagedMenu that we want to bind the "Back"-button on the first page on ourselves
    # the argument is what would be passed as option.value in the callback
    menu.add_back_option('go_back')

    menu.select_callback = some_other_menu_callback
    menu.send(index)

def some_other_menu_callback(menu, index, option):
    if option.value == "go_back":
        # would be only called if we are in the first page of the menu, otherwise behave normally
        return some_menu(index)

    print(option.value)

```

Might not be the best way to do this, but it's just an idea so you know what I mean.

MarkusMeskanen:
We did something similar on [Hero-Wars](https://github.com/Hero-Wars-Dev-Team/Hero-Wars/blob/master/addons/source-python/plugins/hw/menus.py#L72). Basically there's `previous_menu` attribute for PagedMenus, and pressing `Back` on page 1 will open the previous menu instead.

high-voltage:
@MarkusMeskanen This would mean that every menu needs to be in the global scope of the script if the previous_menu attribute calls a PagedMenu instance, doesn't it? 

I also thought about adding an own callback for this option, but the way I did it works best in my opinion. Tell me what you think about it.

Ayuto:
Added with PR #111.

Issue #52, opened 2015-05-18
menus.PagedMenu() close option to button 9. in csgo


high-voltage:
For PagedMenu() I would like to have the buttons Back, Next and Close set to 7, 8 and 9 respectively because the 0 button is not bound by default in csgo (so stupid...).

satoon101:
This has already been planned, but thank you for creating an official issue report for it.

PR #55, opened 2015-05-18
Changed PagedMenu for csgo and added manual back option

high-voltage:
As to the issues #52 and #53 I've made changes and tested them, they work well. Maybe you want to use it :)

Working example code:

```

from events import Event

from players.helpers import index_from_userid

from menus import SimpleMenu, SimpleOption, PagedMenu, PagedOption



@Event
def player_jump(game_event):
    userid = game_event.get_int('userid')
    index = index_from_userid(userid)
    some_menu(index)
    
    
def some_menu(index):
    menu = SimpleMenu()
    menu.append(SimpleOption(1, 'some other menu', 1))
    menu.select_callback = some_menu_callback
    menu.send(index)

def some_menu_callback(menu, index, option):
    if option.value == 1:
        some_other_menu(index)

def some_other_menu(index):
    """
     Added a "manual_back_option" parameter, which is defaulted to None.
     When pressing the "Back" button in the first page of the menu,
     this will be passed as option.value in the callback.
    """
   
    menu = PagedMenu(manual_back_option='go_back')
    
    for x in range(1, 30):
        menu.append(PagedOption(x, x))

    menu.select_callback = some_other_menu_callback
    menu.send(index)

def some_other_menu_callback(menu, index, option):
    if option.value == "go_back":
        return some_menu(index)

    print(option.value)

```

### DIFF
--- a/addons/source-python/packages/source-python/menus/radio.py
+++ b/addons/source-python/packages/source-python/menus/radio.py
@@ -15,6 +15,8 @@ from menus.base import _translate_text
 from menus.queue import _radio_queues
 #   Messages
 from messages import ShowMenu
+#   Core
+from core import SOURCE_ENGINE
 
 
 # =============================================================================
@@ -26,6 +28,21 @@ __all__ = ('PagedRadioMenu',
            'SimpleRadioOption',
            )
 
+# =============================================================================
+# >> GLOBALS
+# =============================================================================
+_footer_select_indexes = {
+    'csgo': {
+        'back': 7,
+        'next': 8,
+        'close': 9
+    },
+    'orangebox': {
+        'back': 8,
+        'next': 9,
+        'close': 0
+    }
+}
 
 # =============================================================================
 # >> CLASSES
@@ -135,7 +152,7 @@ class PagedRadioMenu(SimpleRadioMenu, _PagedMenuBase):
             self, data=None, select_callback=None,
             build_callback=None, description=None,
             title=None, top_seperator='-' * 30, bottom_seperator='-' * 30,
-            fill=True):
+            fill=True, manual_back_option=None):
         """Initialize the PagedRadioMenu instance.
 
         @param <data>:
@@ -181,10 +198,11 @@ class PagedRadioMenu(SimpleRadioMenu, _PagedMenuBase):
         self.top_seperator = top_seperator
         self.bottom_seperator = bottom_seperator
         self.fill = fill
+        self.manual_back_option = manual_back_option
 
     def _get_max_item_count(self):
         """Return the maximum possible item count per page."""
-        return 7
+        return 6 if SOURCE_ENGINE == 'csgo' else 7
 
     def _format_header(self, player_index, page, slots):
         """Prepare the header for the menu.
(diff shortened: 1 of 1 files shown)


high-voltage:
I guess that counts only for the numbers of the buttons (7,8,9 / 8,9,0), what about the manual_back_option parameter?

necavi:
This can't really be data driven, considering it is based on a client-side binding and only affects one game.

satoon101:
The back option (on the first page) could have a menu assigned to it.  Not sure I would call it manual_back_option, more like parent_menu which would default to None.

The buttons is what I was talking about.  And yes, it can certainly be data driven.

high-voltage:
If we have it parent_menu, it should most likely automatically call a menu. But wouldn't this require to have all our menus (or atleast the needed ones) in the global scope or saved in a dict? I think it's the best way to have it like I intended, maybe just not the name "manual_back_option" but I couldn't come up with something that sounds better.

satoon101:
They wouldn't all have to be in the global scope.  When declaring the parent_menu for a child menu, it would just require access to the parent menu's instance.

MarkusMeskanen:
This actually is a "problem" in the current implementation too; say I want to have 5 menus, one "Main Menu" and 4 sub menus. If I have them all in the same file, my main menu must be the last one since it "needs to" (should?) have the other 4 menus accessible in it's constructor. It would be better in the first place if the menu content would somehow be defined in a function instead of a global menu instance. Just my opinion. :)

satoon101:
All menus inherit from list.  This allows you to append options/text to the menu in order.

MarkusMeskanen:
I understand, but I can't see why they couldn't just contain a list and have similar methods like `add` etc.

I think every menu should be subclassed from some kind of base menu, and the build callback and select callback would be methods that override the base menu's methods. This way your menus could basically do anything without having to worry about whether the other menus in global scope are defined before them. At least that's how I'd do it, any reason not to do it this way? :)

satoon101:
Why would it contain a list when it "is" a list?  And you use the "append" method to add items.  Unless I am missing something, I think you already do that with build and select callbacks.  The only thing with them being in the global scope is that they cancel/close automatically when the script unloads.

high-voltage:
Back to the "back" option, let's take my example script above as an example. 
If we would have the "parent_menu" parameter, which requires a menu instance, I would have to make the menu in "def some_menu(index)" in the global scope or save it somewhere else. I personally don't like to do that, and it would complicate everything even more if the parented menu displays player specific information, which could refresh while browsing through the menus.

MarkusMeskanen:
Satoon, the idea would be that you don't need to have the sub menus defined before the main menu (in global scope f.e.).

It really isn't that big of an issue, but here's what I mean: https://gist.github.com/MarkusMeskanen/5301961ee6a384aef768

But probably not worth the time or anything :)

satoon101:
The items are what make up the menu.  It makes perfect sense to me.  Check your link above again, as Ayuto and I have provided further examples.

satoon101:
@high-voltage, if we added it as a parameter, we would also store it as an attribute.  That way you can not pass it on instantiation and set the value later.

Ayuto:
Due to PR #111 I will close this.

PR #63, opened 2015-06-05
Fix CSourcePython::OnQueryCvarValueFinished debug message.

### LINKED ISSUE #62: Incorrect debug message level in CSourcePython::OnQueryCvarValueFinished
A debug message is printed to the console when OnQueryCvarValueFinished is called if you have `level` set to `1` or higher in `core_settings.ini`.

To reproduce:

```python
from engines.server import engine_server
from events import Event
from players.helpers import edict_from_userid

@Event
def player_say(event):
    userid = event.get_int('userid')
    engine_server.start_query_cvar_value(edict_from_userid(userid), 'net_fakejitter')
```
Results in something like:
`Cvar query (cookie: 143, status: 0) - name: net_fakejitter, value: 0`

May 21st release.

jsza:
Fix #62

### DIFF
--- a/src/core/sp_main.cpp
+++ b/src/core/sp_main.cpp
@@ -390,7 +390,7 @@ void CSourcePython::OnQueryCvarValueFinished( QueryCvarCookie_t iCookie, edict_t
 void CSourcePython::FireGameEvent( IGameEvent * event )
 {
 	const char * name = event->GetName();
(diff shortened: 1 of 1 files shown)


satoon101:
While you are at it, can you change the FireGameEvent message to use 4, as well?  Once you do that, I will pull the request.

satoon101:
Thank you.

PR #73, opened 2015-09-30
Fix usermessages not being sent when player classname is changed.

jsza:
It is possible to set the classname keyvalue of a player entity to something other than "player". Some custom maps make use of this in conjunction with filter_activator_class for various things.


Ayuto:
Thanks!

PR #75, opened 2015-10-05
Fix menu close button slot.

jsza:
The slot for the menu close button needs to be 0 instead of 10 for games other than CS:GO/BMS.

Not sure if this is the best way to fix it - let me know if it needs any changes.


Ayuto:
Thanks! We should probably raise an error here: https://github.com/jsza/Source.Python/blob/17d195b71f63fcd4d8dc38b2402d88f9ca02a4fc/addons/source-python/packages/source-python/menus/radio.py#L98

Then we would have noticed that issue earlier. Could you please add that as well?

jsza:
Pushed another commit. Is that what you had in mind?

Ayuto:
Yep, thanks! :)

PR #76, opened 2015-10-28
Added 'event_name' to event's kwargs

MarkusMeskanen:
Event's name is still required when registering the same callback for multiple events:

    @Event('player_spawn', 'player_jump')
    def on_event(**eargs):
        print('Player with userid of {userid} triggered {event_name}.'.format(**eargs))

I chose to use `'event_name'` since it can still be grabbed in the parameters (`'event name'` could not be):

    @Event('player_spawn')
    def on_spawn(event_name, **eargs):
        pass

And simply using `'name'` felt a little ambiguous.

### DIFF
--- a/addons/source-python/packages/source-python/events/hooks.py
+++ b/addons/source-python/packages/source-python/events/hooks.py
@@ -183,6 +183,7 @@ def _pre_game_event(args):
     # TODO: use data or some other means to get the offset
     kwargs = make_object(
         KeyValues, get_object_pointer(game_event).get_pointer(8)).as_dict()
+    kwargs['event_name'] = event_name
 
     # Loop through all callbacks in the pre-event's list
     for callback in pre_event_manager[event_name]:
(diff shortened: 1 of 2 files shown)


MarkusMeskanen:
`GameEvent` objects also have `is_local()` and `is_reliable()` which are both ignored by the `KeyValues`, but I left them out as I wasn't sure if they're relevant enough to be added to the dict.

satoon101:
Good point.  Thank you.  I'm not sure about the other two, as well.

We "could" also pass the GameEvent object itself as a kwarg, which would fix the issue I noticed last night where we cannot directly duplicate the event anymore.  Not sure how widely used that would be, though.

MarkusMeskanen:
I also thought about passing the `GameEvent` object as a kwarg, but I found very little use to it. What exactly do you mean by "directly duplicate the event"?

satoon101:
events.manager.game_event_manager has a method to duplicate an event.  It requires the GameEvent object to be passed.
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/events/events_wrap.cpp#L244

It was mentioned in this thread of your's previously:
http://forums.sourcepython.com/showthread.php?727

However, as I said, I'm not sure how widely used it would even be.  Creating a custom event instead, and passing in the kwargs would be just as good an option.

PR #80, opened 2015-11-08
Updated the weapon system to not use extra classes

Mahi:
`Player` class used multiple classes to achieve stuff like `get_weapon_color()`, which internally received the player's active weapon and returned it's `color` attribute. However, there was no way to get the player's weapon itself (although it was used in some of the methods), so I decided to rewrite the weapon system. I completely got rid of the old system which used multiple classes and multi-inheritance for no apparent reason, and instead I implemented methods to the `Player` class for receiving the player's actual weapon objects.

If you now want to change player's ammo or his weapon's color, you can simply get the `Weapon` instance (and operate on it instead of the player) using `player.active_weapon`, `player.primary_weapon`, and `player.secondary_weapon` properties, as well as the `player.get_weapons()` method which will return a generator of all player's weapons.


MarkusMeskanen:
Also, any thoughts on setters, would they be useful?

The `active_weapon`'s setter could either equip the player with the provided weapon and drop/delete his existing weapon on that slot (if any), or only work if the player already has the weapon equipped (f.e. `player.active_weapon = player.secondary_weapon` to force him to switch to his pistol). The setters for primary and secondary weapons would probably equip him with the new weapons.

Is any of this necessary, or are the getters enough?

necavi:
I think that that kind of thing really needs to be explicit, using functions like player.changeweapon() or something similar.

MarkusMeskanen:
Yeah, you might be right. Thre getters should still stay, though.

necavi:
I completely agree, they'll be very useful.

invincibleqc:
Your implementation is slow, very slow. To find the player's weapons, the old one was looping directly on the player's props (which we already know the ownership) while you are looping through ALL the existing weapons on map to then compare ownership - thousands of extra loops on large servers.

satoon101:
While I do agree with getting rid of a lot of the methods we currently have and just returning the Weapon instance, I do not like the fact that your implementation removes all game specific and projectile functionality.

And @invincibleqc is right about the looping.  We have had this discussion before on the forums, though I cannot seem to find it right now.

I have already been working on an update for this anyway, maybe I'll try to get it done and committed today.  I do like your active weapon change, though.  That's not something I had thought of previously.

MarkusMeskanen:
@satoon101 I still don't understand everything about the engine and SP, what's are the projectiles even used for?

What comes to the looping, I think we're talking about milliseconds here, but that's still a valid point. I should've probably used the old method of going through the player's props.

Ayuto:
I actually like the proposed changes. Though, they are now conflicting with our player_weapons_update branch.

Regarding the performance: Looping through the player's weapons will always have the same performance with the old method. The new one will vary depeding on the number of weapons on the map.

I also like that you moved the weapon methods to the Player class. I'm not a fan of splitting a class into multiple base classes that are splitted over several files just to keep files small.

One downside of this implementation is that you removed quite a few methods like get_flashbang_count etc. They are hidden in a meta class that creates them. I'm not sure if we really need them, but we should definitely get rid of the meta class. That will make the code much more readable, flexible and probably faster (negligible, because it's only initialization code).

satoon101:
> Regarding the performance: Looping through the player's weapons will always have the same performance with the old method. The new one will vary depeding on the number of weapons on the map.

That last statement is not entirely accurate.  You still iterate over the max entities, not just all weapons.  The current implementation only iterates over the size of m_hMyWeapons, which is typically either 48 or 64 (much fewer than 2048).

satoon101:
> One downside of this implementation is that you removed quite a few methods like get_flashbang_count etc. They are hidden in a meta class that creates them. I'm not sure if we really need them, but we should definitely get rid of the meta class. That will make the code much more readable, flexible and probably faster (negligible, because it's only initialization code).

You're right about that, I will go ahead and remove the meta class and just right in the methods.


> I also like that you moved the weapon methods to the Player class. I'm not a fan of splitting a class into multiple base classes that are splitted over several files just to keep files small.

I disagree with this.  It isn't about keeping files small, it's about separation of ideas.  We have a players.weapons sub-package to house all player specific weapon functionality.  The primary class is then inherited by Player to expose the functionality to that class.

Ayuto:
If it's not about keeping files small, we can simply use comments to separate those methods just like we do with imports, constants, global variables, classes, functions, etc.
```
class A(object):
    # =========================================================================
    # >> WEAPON FUNCTIONALITY
    # =========================================================================
    def get_weapons(self, ...):
        pass
        
    @property
    def active_weapon(self):
        pass
        
    @property
    def primary_weapon(self):
        pass
        
    @property
    def secondary_weapon(self):
        pass
        
        
    # =========================================================================
    # >> MUTE FUNCTIONALITY
    # =========================================================================
    def mute(self, ...):
        pass
        
    def unmute(self, ...):
        pass
        
    def is_muted(self, ...):
        pass
```

satoon101:
I understand that, but we have a players.weapons sub-package anyway for multiple purposes, not just the basic functionality.  This will likely even expand in the future when we start supporting more games further.

Ayuto:
As far as I understood the players.weapons package it's only adding methods like get_flashbang_count (supported by CS:S and CS:GO) and get_decoy_count (only supported by CS:GO). Maybe we should just provide a get_projectile_count that accepts a projectile name.

Or if we want to provide wrappers for that method they should be available on all games, but raise a proper error if the game is not supported (e.g. decoy on CS:S). Currently, it's only raising an AttributeError (afaik). Checking for the availibility of a weapon in a game can be done in 2-3 lines and doesn't need the players.weapons package.

Pseudo example:
```
    def _get_projectile_count(self, projectile_name):
        weapon_class = weapon_manager[projectile_name]
        if weapon_class is None:
            raise NotImplementedError(
                'Weapon "%s" is not supported by this game.'.format(
                    projectile_name))
            
        # Code...
            
    def get_decoy_count(self):
        return self._get_projectile_count('weapon_decoy')
```

satoon101:
Currently it adds the projectile and c4 functionalities.  It is setup to be expanded upon once we start supporting more games.

And, it does raise an AttributeError in non-supported games, because it is not added as a method at all for games the method does not work on.

MarkusMeskanen:
Even if the `players.weapons` package is needed anyways, I still agree with Ayuto that the `_PlayerWeapons` class should be removed and the content should be in the `Player` class.
Creating a separate class which does nothing but provides a few methods for an other class via inheritance is one weird design pattern which simply uses unnecessary classes and creates a more complex structure. As far as I can see, there's absolutely no benefit of using such structure. Correct me if I'm wrong though.

Ayuto:
> Currently it adds the projectile and c4 functionalities. It is setup to be expanded upon once we start supporting more games.

If it's expanding it still belongs to the Player class.

> And, it does raise an AttributeError in non-supported games, because it is not added as a method at all for games the method does not work on.

What I was trying to say is that one of our original ideas was that SP provides every method for every game, but raises a descriptive error if the game is not supported. And this could be done easily even if all methods were in the Player class (it only takes 2-3 lines of code in total).

satoon101:
We cannot provide the NotImplementedError methods directly into the Player class itself, as those will override any that we inherit from.  I have made some changes according to all of this here:
https://github.com/Source-Python-Dev-Team/Source.Python/commit/55c99cdf23369d963d107cc702171f8c49c988f0

Ayuto:
I like the change, but I think we should also rename generators like "weapons" and "weapon_indexes". We usually start methods with a verb. In this case they are getters, so they should start with "get_" to avoid confusion with attributes/properties.

Which methods would be overwritten? I think we can implement this easily with this code.
```
    @staticmethod
    def _get_weapon_class(projectile_name):
        weapon_class = weapon_manager[projectile_name]
        if weapon_class is None:
            raise NotImplementedError(
                'Weapon "%s" is not supported for this game.'.format(
                    projectile_name))

        return weapon_class

    def get_projectile_ammo(self, projectile_name):
        return self.get_property_int(
            weapon_manager.ammoprop +
            '%03d' % self._get_weapon_class(projectile_name).ammoprop)

    def set_projectile_ammo(self, projectile_name, value):
        return self.set_property_int(
            weapon_manager.ammoprop +
            '%03d' % self._get_weapon_class(projectile_name).ammoprop, value)

    def get_hegrenade_ammo(self):
        return self.get_projectile_ammo('hegrenade')
        
    def set_hegrenade_ammo(self, value):
        return self.set_projectile_ammo('hegrenade', value)
        
    # And so on with all projectiles...
```
I'm not even sure if we want to keep the get_<projectile>_ammo/set_<projectile>_ammo wrappers if we provide a generic get_projectile_ammo/set_projectile_ammo method. The benefit of having them would be that we can provide properties (``player.flashbang_ammo = 2``).

Moreover, I think we should remove the projectile generators like "flashbang_indexes". The way of getting the projectile indexes doesn't differ from other weapons. And I don't think we want to start adding "m4a1_indexes" etc.

satoon101:
I can agree with the <type>_indexes.  I already previously removed the get_<type>_index_list methods and considered removing the <type>_indexes, as well.  Part of the issue is that incendiary isn't handled the same as other projectiles in CS:GO, since both weapon_molotov and weapon_incgrenade utilize the same ammoprop.  When either is thrown, it is molotov_projectile entity and both kill with the inferno entity.  For those weapons, we cannot rely on just using the molotov or incgrenade name.  We have to instead use an is_filter for incendiary.

The reason I did it the way in the newest commit is for that purpose, as well as any future games that might use the same names but need to be handled differently.  For instance, what if another game uses both incgrenade and molotov (or really just 2 incendiary weapon types), but they are handled separately instead of using the same ammoprop?  We then would need to override the current method for that specific game.  The method you propose would not work in that instance, as the built-in method in Player would not be overridden by any game-specific method.

Ayuto:
> Part of the issue is that incendiary isn't handled the same as other projectiles in CS:GO, since both weapon_molotov and weapon_incgrenade utilize the same ammoprop.

Ah, I missed that. Thanks! I think in that case it makes sense to keep the generators. But we should still rename them.

> The reason I did it the way in the newest commit is for that purpose, as well as any future games that might use the same names but need to be handled differently.

One could argue this could also happen with other implementations. E.g. the view_vector calculation could differ from game to game. That would mean we need to move everything into game specific files. Imo, we should keep everything within a shared class as long as it's possible.

satoon101:
Alright, I can be on board with that.  I realized after I posted that that we already do have an issue when it comes to incendiary.  L4D2 has both molotov and incendiary ammo boxes.  Neither currently utilizes the incendiary tag, though both should.  So, at the very least, incendiary methods should be handled in the game-specific modules.  I can make these changes tonight when I get home from work.

satoon101:
We have merged the player_weapons_update2 branch into master, so this pull request is no longer necessary.

PR #87, opened 2015-11-30
It it "LocalActiveWeaponData" according to CS:S.

KirillMysnik:
Difficult to notice, took a while to figure out

### DIFF
--- a/addons/source-python/data/source-python/entities/CBaseCombatWeapon.ini
+++ b/addons/source-python/data/source-python/entities/CBaseCombatWeapon.ini
@@ -5,8 +5,8 @@
 
 [property]
 
-    next_attack = LocalWeaponActiveData.m_flNextPrimaryAttack
-    next_secondary_fire_attack = LocalWeaponActiveData.m_flNextSecondaryAttack
+    next_attack = LocalActiveWeaponData.m_flNextPrimaryAttack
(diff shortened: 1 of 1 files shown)


satoon101:
Actually, using LocalActiveWeaponData is correct for the networked property, while without is the non-networked one.  That is true for all engines/games according to my entity_properties tests:
https://www.dropbox.com/s/vdao6lcsr3fs9e7/CBaseCombatWeapon.txt?dl=0

Either technically works, but we should be using the networked one so that so that [Edict.state_changed](https://github.com/Source-Python-Dev-Team/Source.Python/blob/41588a476262861337d6678d780d3a02bd073d1a/addons/source-python/packages/source-python/entities/classes.py#L571) will be called.

Thank you!

KirillMysnik:
Variant with "LocalWeaponActiveData" didn't work for me, the attributes `next_attack` and `next_secondary_fire_attack` were not getting created. I was getting AttributeError.

KirillMysnik:
Yes, the point is that "Active" and "Weapon" were mixed in the previous version, not that it was just simply without the `LocalActiveWeaponData` prefix.

satoon101:
Oh yeah, sorry, was on phone and didn't realize that.  Thanks again.

PR #90, opened 2015-12-03
Updated entities to use object.__setattr__()

Mahi:


Ayuto:
I'm not sure if we really want to use object.__setattr__ instead of using super(), but I will merge it for now. Thanks for the PR!

Mahi:
Yup, that's up to you guys how you eventually to be done. I find it better (explicit is better than implicit!) and clearer, but that's just me and I'm happy with any fix at all, so thanks for the merge!

satoon101:
I get the explicit/implicit argument, but I have always read that you should use super in case of a change in inheritance.  I understand that that really wouldn't be an issue in this instance, but I would prefer to remain consistent with this.

Mahi:
> I get the explicit/implicit argument, but I have always read that you should use super in case of a change in inheritance.

As far as I'm aware, this simply means that if you don't know how to set the attribute, ask your base class to set it. However, since you were already explicitly skipping the base class to get to `object.__setattr__` (using `super(Entity, self)` in `Player` class instead of `super(Player, self)`), I think that negatiates the whole idea of using `super()` so it's just better to explicitly use `object.__settattr__`. This is not true inside of the `Entity` class since you're not skipping anything, but even in `Entity` you'd probably skip the base class if it had any weird logic in its `__setattr__` to get straight to the `object.__setattr__`. I could be wrong though, but that's how I see it, you make the final decision.

PR #94, opened 2015-12-06
Kwargs and return values to filters and hooks' __call__

Mahi:
Should be quite self-explanatory. Even if they're not currently used by anything, this improves extendability and if someone wants to use kwargs or the return value, let them.


Ayuto:
Thanks!

Mahi:
Glad to help!

PR #104, opened 2015-12-26
Removed unneccessary equals sign in effects data

Predz:
Found this by accident whilst testing the new effect changes.


Predz:
I also seem to be having trouble with getting life times to work with effects. They always seem to last for ever. Not to sure if we have the wrong property being used in the data.

invincibleqc:
Thanks!

PR #106, opened 2016-01-04
Updated entities.specials to support new conversion functions

KirillMysnik:
By new conversion functions I mean the ones that don't provide optional 'raise_exception' keyword argument any more.


Ayuto:
Thanks!

PR #111, opened 2016-01-26
Added parent_menu to PagedESCMenu and PagedRadioMenu

ThaPwned:


ThaPwned:
Yes, thank you! 'Em copy and paste mistakes...

Ayuto:
Thanks for the PR!

Mahi:
Not like it really matters, but there are two of these lines:

    if not page_index and self.parent_menu is not None:

Which should probably explicitly check for `page_index == 0` instead of `not page_index`.

Ayuto:
True, but if we are already nit picking, we should use a constant like FIRST_PAGE. :P

Mahi:
I believe you missed my point here. It's not just a semantic change, changing `not page` to `page_index == 0` actually changes the behaviour of the interpreter. If someone were to do something wrong and accidentally managed to set the `page_index` to an empty list, the `not page_index` check would evaluate as `True`.

satoon101:
Well, playing devil's advocate, if we believe "we're all adults", then if someone decides to change page_index to anything other than an integer, they deserve to have issues.

Ayuto:
> I believe you missed my point here.

No, I got your point. That's why I agreed, but this is really just a minor detail. 

Btw. you can't set the page index to something else than an integer. set_player_page() validates the input. But even if you set it to an empty list, you will run into errors when the menu is refreshed.

PR #121, opened 2016-05-01
Fixed a typo in bms/CDynamicProp.ini

KirillMysnik:


Ayuto:
Thanks for the PR!

Btw. you can also update a fork without that additional commit. You just need to rebase your fork:
http://stackoverflow.com/a/7244456

KirillMysnik:
Thanks for the tip! Will try to do that next time. I'm still noobish at the git stuff and only work with it via GUI of PyCharm.

Issue #97, opened 2015-12-09
<Sound object>.duration always returns 0


Ayuto:
Reference:
http://forums.sourcepython.com/showthread.php?940-Plugins-converts-from-Eventscripts-to-Source-Python&p=6441&viewfull=1#post6441

A possible solution is to add a sound library like [mutagen](https://mutagen.readthedocs.org/en/latest/). This would allow us to add more attributes to the Sound class.

Question:
What do we do with sound files packed in VPK files? Actually, this is not just about sound files, but any file within a VPK file. What if we want to call a function that requires a file path? Should we create a temporary file?

KirillMysnik:
Real-time VPK unpacking?

How about manually hashing all stock game VPKs and just creating corresponding tables to them which contain information like
<sound path> - duration
?

And if the server uses a custom VPK - well, then unpack VPK either in real time on request (slow?) or unpack and create such tables on server startup.

Ayuto:
Actually, reading a file from a VPK file seems to be fast and it's probably much faster if we create the file-like object with the functionality provided by the SDK (because it's implemented in C++). If someone needs a name on the file system, we could provide a named temporary file (there is a Python library for that).

I don't want a hash table, because that only fixes the duration issue and not accessing packed files.

dsezen:
No need for special libraries. IEngineTool contains exactly what you need ;)

satoon101:
This should all be fixed as of this commit:
https://github.com/Source-Python-Dev-Team/Source.Python/commit/bc4b1ded4110d8d105a36f81721686b204510b5b

Ayuto:
The reason for reopening this issue is that the interface is not available on Linux.

Killer0561:
If this interface for Linux ever be available? This topic is now long open without a solution, unfortunately.

satoon101:
@Killer0561 - That interface will never be available for Linux (or at least very highly unlikely).  We are working on a solution.  The most recent commit should allow access to the duration for all wav, mp3, and ogg files that are directly on the server.  We still have to figure out how we want to get the duration for sounds in vpk files.

satoon101:
The newest solution seems to work fine, but I would like more people to do more testing on it to verify.  It now allows for getting the sound duration from sounds within vpk files.  My test script on CS:S was:
```python
from engines.sound import Sound
from stringtables import string_tables

sounds = [
    'hostage/hpain/hpain1.wav',
    'player/headshot1.wav',
[...]

PR #122, opened 2016-05-02
Initial wrap of useful functions from the IEngineTool interface.

dsezen:
Hi,

This change wraps the IEngineTool interface which contains some very useful functionality. Currently, I've only wrapped the stuff that isn't covered by the IVEngineServer wrapper. I've also skipped stuff that only works client-side. This should also help solve issue #97.


satoon101:
Thank you for this.  I just tested and it works great!!  However, you did forget one thing.  You need to add the new file to the CMakeLists.txt.

dsezen:
Done.

satoon101:
Thank you!!

PR #126, opened 2016-05-17
Add weakref-based AutoUnload class to allow instances to be garbage collected

### LINKED ISSUE #125: Instances of AutoUnload subclasses can't get garbage collected
This behaviour is caused, because the instances are added to ``module_instances``:
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/core/__init__.py#L89

However, we don't want all instances to get garbage collected. Delay instances for example should be removed from ``module_instances`` when the delay has finished while Event instances should stay in ``module_instances`` until the plugin has been unloaded.

jsza:
Fixes #125

### DIFF
--- a/addons/source-python/packages/source-python/core/__init__.py
+++ b/addons/source-python/packages/source-python/core/__init__.py
@@ -19,6 +19,8 @@ from path import Path
 from platform import system
 #   Sys
 import sys
+#   Weakref
+import weakref
 
 # Site-Packages Imports
 #   ConfigObj
@@ -85,11 +87,15 @@ class AutoUnload(object):
         # Get the calling module
         caller = getmodule(stack()[1][0])
 
-        # Set the _calling_module attribute for the instance
-        _module_instances[caller.__name__].append(self)
+        # Call class-specific logic for adding the instance.
+        self._add_instance(caller.__name__)
 
         # Return the instance
         return self
+
+    def _add_instance(self, caller):
+        """Add the instance to _module_instances."""
+        _module_instances[caller].append(self)
 
     def _unload_instance(self):
         """Base _unload_instance implementation."""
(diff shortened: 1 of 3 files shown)


jsza:
I originally planned to add this functionality to AutoUnload itself, but since no references are stored for Events and such (causing them to be garbage collected instantly), this seemed like the easier approach.

addons/source-python/packages/source-python/core/__init__.py
```diff
@@ -101,6 +107,36 @@ def _unload_instance(self):
...
 
 
+class WeakAutoUnload(AutoUnload):
+    """Subclass of AutoUnload used to unload specific instances when all
+    references to the instance are deleted.
+    """
+
+    def new(cls, *args, **kwargs):
+        self = super().__new__(cls)
+        self.___unloaded = False
```
Mahi: What's with the triple underscore here? I don't think this follows any convention in Python

Ayuto:
Thanks for the PR! But this does not fully solve the problem, because ``_module_instances`` will still continue to grow. I should have mentioned that in the issue. A better approach would be to use two ``_module_instances`` dicts as class attributes. The AutoUnload class would continue using ``defaultdict(list)`` and the WeakAutoUnload class would use ``defaultdict(WeakValueDictionary)``. Obviously, WeakSet would be the preferred choice, but that would require us to make all WeakAutoUnload subclasses hashable. So, WeakValueDictionary is a perfect workaround (``id(<WeakAutoUnload object>)`` would be the key and ``<WeakAutoUnload object>`` would be the value).

addons/source-python/packages/source-python/core/__init__.py
jsza: Yeah, thanks, it seems this should have been 2 underscores as per https://www.python.org/dev/peps/pep-0008/ (to avoid naming collisions for subclasses).

jsza:
Ah, thanks for the reply Ayuto! That's a very elegant solution. Is it okay if I make those changes here or would a new branch be preferred?

Ayuto:
Just make them here. :)

addons/source-python/packages/source-python/core/__init__.py
Ayuto: Somewhere we defined that we don't want to use super-private attributes. That's why we never use 2 underscores in SP. It seems like that point is missing in the wiki.

Mahi: @Ayuto Those aren't super-private attributes, they're to avoid name collision in subclasses like @jsza said. [Here](https://www.youtube.com/watch?v=HTLu2DFOdTg&t=33m8s)'s a good talk about it from Raymond Hettinger.

jsza:
Not sure if GitHub notifies for new commits on PRs, but I've pushed the changes above.

Ayuto:
Thanks for the update! But there are still two things:

1. WeakAutoUnload instances shouldn't call ``_unload_instance`` when they get garbage collected. If an instance whishes to do stuff when it gets garbage collected, it should implement ``__del__`` on its own. ``_unload_instance`` is only for for the plugin manager. This also makes sure that ``_unload_instance`` is only called once (when the plugin has been unloaded) or never (when it gets garbage collected before the plugin unloads). Thus, all the ``__new__``, ``__del__``, ``_instance_unloaded`` and ``_unload_instance override`` stuff is redundant.
2. I'm not fully sure if you have fully tested the code. Otherwise you should have noticed the ``__new__`` spelling error.

jsza:
@Ayuto Thank you, seems I didn't test it well enough. I'll be be more careful with contributions in future.

jsza:
Aaand I just noticed the doc string was wrong. Oops.

Ayuto:
Thank you! Now everything seems to be fine. :)

PR #134, opened 2016-06-29
Added decorators for adding callbacks to menus

Mahi:
This allows one to define the menu before its callbacks while keeping the code clean.

### DIFF
--- a/addons/source-python/packages/source-python/menus/base.py
+++ b/addons/source-python/packages/source-python/menus/base.py
@@ -201,6 +201,35 @@ class _BaseMenu(AutoUnload, list):
         """
         raise NotImplementedError
 
+    def register_select_callback(self, callback):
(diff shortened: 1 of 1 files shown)


Ayuto:
This just moves a single line...
```python
def my_callback(...):
    pass
menu.select_callback = my_callback
```
```python
menu.register_select_callback
def my_callback(...)
    pass
```
I would rather update the menus package to support overrideable methods instead of callbacks.

Mahi:
But it's much more beautiful and clear (think of a very long function, that `menu.select_callback = my_callback` will get lost in the clutter quite easily, unless you separate it with an empty line or two) what you're doing. I find it much more simpler at the cost of nothing. But you guys decide! :)

> I would rather update the menus package to support overrideable methods instead of callbacks.

Me too, but until someone has enough time and patience to do that...

PR #136, opened 2016-07-08
Multiple improvements / bug fixes to player_weapons_update2 branch

KirillMysnik:
Resolved confusion between clip/secondary_fire_clip attributes defined in data files and get_clip/set_clip/... methods in `_entity/__init__.py`
Special thanks to @Mahi for consulting this PR

### DIFF
--- a/addons/source-python/data/source-python/entities/CBaseCombatWeapon.ini
+++ b/addons/source-python/data/source-python/entities/CBaseCombatWeapon.ini
@@ -10,6 +10,6 @@
     ammoprop = LocalWeaponData.m_iPrimaryAmmoType
     secondary_fire_ammoprop = LocalWeaponData.m_iSecondaryAmmoType
     owner_handle = m_hOwner
-    clip = m_iClip1
-    secondary_fire_clip = m_iClip2
+    _clip = m_iClip1
+    _secondary_fire_clip = m_iClip2
     flip_view_model = LocalWeaponData.m_bFlipViewModel
--- a/addons/source-python/docs/source-python/source/contributing/reporting-a-bug.rst
+++ b/addons/source-python/docs/source-python/source/contributing/reporting-a-bug.rst
@@ -10,5 +10,6 @@ a list of points you should always try to provide:
 
 1. A precise description of the problem.
 2. A tested code snipped to reproduce the issue.
+3. The full traceback of the exception (if an exception was raised).
 
 Thank you!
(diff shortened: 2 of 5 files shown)


KirillMysnik:
Or, as I have seen somewhere in SP, we can keep data file untouched and instead go with something like this
```python
clip = self.__getattr__('clip')
return clip if clip != -1 else 0
```

Also see [issue #137](https://github.com/Source-Python-Dev-Team/Source.Python/issues/137)

KirillMysnik:
Also commited [this](https://github.com/Source-Python-Dev-Team/Source.Python/pull/136/commits/5e617ca6e29947394f0e3a8febc9a407619472ff) - not sure how to do a separate PR

KirillMysnik:
Okay, tried to do a separate PR and successfully failed.

Anyways, as me and @Ayuto discovered, CS:GO's ammo property is `ushort`, not `int`, thus simple `player.get_ammo` was raising exceptions on master. But as Ayuto said, that's not an issue for CS:GO, because it uses `primary_ammo_count` (weapon's property) instead of player-related ammo property. and that was already implemented in player_weapons_update2 branch.
What we both forgot is [this](https://github.com/Source-Python-Dev-Team/Source.Python/blob/player_weapons_update2/addons/source-python/packages/source-python/weapons/_entity/csgo/__init__.py#L21):
```python
if 'grenade' in weapon_manager[self.classname].tags:
    return super(Weapon, self).get_ammo()
```
and [this](https://github.com/Source-Python-Dev-Team/Source.Python/blob/player_weapons_update2/addons/source-python/packages/source-python/weapons/_entity/csgo/__init__.py#L29):
```python
if 'grenade' in weapon_manager[self.classname].tags:
        super(Weapon, self).set_ammo(value)
        return
```
As you can see, in case of CS:GO, it still uses the old get_ammo/set_ammo which in turn utilize wrong property type. [So I replaced `super()` calls with the improved code of original get_ammo/set_ammo.](https://github.com/Source-Python-Dev-Team/Source.Python/pull/136/commits/3af4907a5c1c743770c956d4602f55b530092fa8)

KirillMysnik:
To anyone who would be handling this PR, please give me directions for fixing #137 - I might just as well include the fix in this PR. I'm not sure if I should go with the solution I proposed in #137 itself, or to only apply the solution in `weapons/_entity/csgo/__init__.py`, making it specific for CS:GO

Ayuto:
Thank you!

PR #139, opened 2016-07-09
Added StreamSound class

KirillMysnik:
This PR adds a StreamSound class that allows playing custom sounds from outside of music folder on CS:GO. The issue is very well described [here](https://wiki.alliedmods.net/CSGO_Quirks#Playing_Custom_Sounds).

Inheritance:
`Sound(_BaseSound)`
`StreamSound(_BaseSound)`

Sample code:
```python
from engines.sound import Sound, StreamSound

my_sound = Sound("some/builtin/sound.wav")
my_stream_sound = StreamSound("my/custom/sound.mp3", download=True)
```

See also [http://forums.sourcepython.com/viewtopic.php?f=31&p=8095](http://forums.sourcepython.com/viewtopic.php?f=31&p=8095)
Thanks to @Mahi and @invincibleqc for the help


Mahi:
I think this is the best solution, nice structure you came up with :)

invincibleqc:
Thank you! However, I noted a regression in the `play` method. The `recipients` argument should be `*recipients` and be passed back to the `RecipientFilter(*recipients)` class (like the current `Sound.play` method does). Its constructor is internally doing proper checkings so you can either pass indexes, iterables of indexes, or RecipientFilter instances that get merged to the object or simply add all players when no arguments are passed (see [RecipientFilter.update](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/filters/recipients.py#L77) for more info).

Also, I don't think that creating a module for each class is the way to go. This could be all done into `sound.py` directly, in my opinion.

KirillMysnik:
Okay, I'll move it to sound.py
What about some Sound/StreamSound factory? Currently plugin creator will need to do his own checks whether the game is CS:GO or not. That's repetitive, plus the list of such games may grow (I suspect Portal 2 and Left 4 Dead 2 to be such games too), so there will be a need to fix these checks in all plugins that exist.
On the other hand, I'm not sure what this factory should look like. Simple
```python
def find_best_sound_class():
    if GAME_NAME in ("csgo",):
        return StreamSound
    return Sound
```
?
Or something that will also check the sample for coming from `music/` directory? Something that will also check if the sample is an .mp3 or .wav?
That's what was discussed in the forum thread, but this time I want to make a factory that would do such checks, while everybody is still able to import either `Sound` or `StreamSound` directly.

PR #143, opened 2016-09-04
Spanish lang added

Hackmastr:
Will soon update the other files


satoon101:
Awesome, thank you!  We will wait until you have all of the files translated before merging, so just commit them to the same branch of your repository so they can be a part of the same pull request.

Ayuto:
The auth_strings.ini file is actually just a relict of the old auth module. I have now removed it completely.

PR #147, opened 2016-09-23
Tick update

Mahi:
Updated `Delay` to use normal arguments for `args` and `kwargs` instead of `*args` and `**kwargs` for the arguments that get forwarded to the delayed callback. Also changed `cancel_on_map_end` to a keyword argument.

### DIFF
--- a/addons/source-python/packages/source-python/listeners/tick.py
+++ b/addons/source-python/packages/source-python/listeners/tick.py
@@ -109,19 +109,17 @@ _delay_manager = _DelayManager()
 class Delay(WeakAutoUnload):
     """Execute a callback after a given delay."""
 
-    def __init__(
-        self, delay, callback, cancel_on_map_end=False, *args, **kwargs
-    ):
+    def __init__(self, delay, callback, cancel_on_map_end=False, args=(), kwargs={}):
         """Initialize the delay.
 
         :param float delay: The delay in seconds.
         :param callback: A callable object that should be called after the
             delay expired.
+        :param tuple args: Arguments that should be passed to the callback.
+        :param dict kwargs: Keyword arguments that should be passed to the
+            callback.
         :param bool cancel_on_map_end: Whether or not to cancel the delay at
             the end of the map.
-        :param args: Arguments that should be passed to the callback.
-        :param kwargs: Keyword arguments that should be passed to the
-            callback.
         :raises ValueError: If the given callback is not callable.
         """
         if not callable(callback):
@@ -132,8 +130,8 @@ class Delay(WeakAutoUnload):
         self.exec_time = self._start_time + delay
         self.callback = callback
         self.cancel_on_map_end = cancel_on_map_end
-        self.args = args
-        self.kwargs = kwargs
+        self.args = tuple(args)
+        self.kwargs = dict(kwargs)
         _delay_manager.add(self)
 
     def __lt__(self, other):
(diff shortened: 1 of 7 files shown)


addons/source-python/packages/source-python/listeners/tick.py
```diff
@@ -109,19 +109,17 @@ def add(self, delay):
 class Delay(WeakAutoUnload):
     """Execute a callback after a given delay."""
 
-    def __init__(
-        self, delay, callback, cancel_on_map_end=False, *args, **kwargs
-    ):
+    def __init__(self, delay, callback, cancel_on_map_end=False, args=(), kwargs={}):
```
satoon101: I have already made most of the changes and pushed them, though I did miss the doc strings.  However, you should never declare default values for arguments as mutable types.  I guess the tuple one is fine, but I used None in my implementation.  But a dictionary should never be used, since it can be updated and cause issues.

Mahi: This is only true if I store the used value though, in my case I never store it so it's fine and I find it more clear that it's supposed to be a dict.

satoon101: It's still very bad practice to do so, and will easily lead other people who read your code down the wrong path.

Mahi: Why is it a bad practice though? I do agree that it's a minor downside that someone else might copy paste my code and thus have a bug in their code, but I'm not going to write my code based on what someone else might copy paste from it.

satoon101: You do realize that threading itself uses this better practice, right?

https://github.com/python/cpython/blob/master/Lib/threading.py#L758

https://github.com/python/cpython/blob/master/Lib/threading.py#L781#L782

https://github.com/python/cpython/blob/master/Lib/threading.py#L786

https://github.com/python/cpython/commit/6fe467c15d15a92fdc843af688363a96060a3554

Code is more often read than written.  Someone could very easily mistake your implementation and it would throw them off.

satoon101:
This request is no longer needed as the changes have already been made.  Thank you, though, for the effort.

addons/source-python/packages/source-python/listeners/tick.py
Mahi: "Better practice" as in your favourite? ^^ I do realize that threading uses `kwargs=None`, and there's absolutely nothing wrong with that approach! I even said it's fine:

> I find it more clear to use a dictionary directly than `None`, not that there's anything wrong with `None` either.

Tbh you're just claiming that it's "better" practice and my practice is bad, yet you have no arguments other than "someone might get confused"... Someone could very easily mistake any implementation, mine works and is perfectly valid so it's just a matter of opinion. I do know that `None` is used more often, but it's highly due to not having to convert the arguments afterwards in the method with `self.x = dict(x)`.

satoon101: You should probably do some research then, cause EVERY instance I have found with Google searching points to this.  And I have done a lot of searching on it in the past.  This argument is getting really dumb and unnecessary...

satoon101: This is just one of many examples of searching for "python dangerous default value", which is what mutable types as argument defaults are called, returns.
http://stackoverflow.com/questions/9526465/best-practice-for-setting-the-default-value-of-a-parameter-thats-supposed-to-be

PR #156, opened 2016-10-08
Tick Repeat/Delay updates

satoon101:
- Changes to Delay
  - 2 new properties
    - time_remaining
      - Return the amount of time remaining in the delay
    - time elapsed
      - Return the amount of time that has elapsed since the delay started
  - Fixed doc-string to correctly show `float` instead of `int` for `delay` in seconds
  - Updated `running` property to show return type
- Changes to Repeat
  - Changed name from TickRepeat to Repeat to match naming scheme of Delay class
  - Added doc-strings to methods/properties
  - Added argument `execute_on_start` to the `start` method to denote whether or not to execute the callback when starting the repeat
  - Renamed some properties to better define what they return
    - remaining -> remaining_loops
    - count -> loop_count
    - limit -> loop_limit
    - timeleft -> time_left
    - elapsed -> time_elapsed
  - Added 2 properties to return values from the current Delay
    - delay_time_remaining
    - delay_time_elapsed
- Changes to both Repeat and Delay
  - Changed args to be a tuple instead of using *args
  - Changed kwargs to be a dictionary instead of using **kwargs
  - Added an optional keyword argument / attribute cancel_on_level_end to denote whether to cancel the Repeat/Delay when the level changes
- Renamed TickRepeatStatus to RepeatStatus to match naming scheme


satoon101:
I have tested all of this and it seems to work fine.  I would like it if at least one more person could verify before merging this into the master branch.

Ayuto:
Looks good, but there is one issue and one thing I would change.

1. If you have a repeat with ``cancel_on_level_end=True`` and you change the map, the repeat will be stopped, but its status remains in the old state (e.g. ``RUNNING`` if it was running before).
2. ``math.inf`` should be used to handle infinite repeats.

satoon101:
Good catch!  I will fix that bug tonight, and implement the infinite change, as well.

maduxthemad:
you could also try to preserve nominal synonymity - the following is somewhat my preference: 

> remaining -> **loop_count_remaining** _(remaining_loops)_
> elapsed -> **loop_count_elapsed** _--- this doesnt exist, but loop_count_elapsed could be added too ---_ 

> count -> loop_count
> limit -> loop_limit

> timeleft -> time_**remaining** _(time_left)_
> elapsed -> time_elapsed
> 
> delay_time_remaining
> delay_time_elapsed

satoon101:
count is the current number of loops elapsed.  I agree with those changes (though slightly different):
> loops_remaining
loops_elapsed

> total_time_remaining
total_time_elapsed

> delay_time_remaining
delay_time_elapsed

> total_loops
total_time

satoon101:
I believe I have fixed everything except for this:
> 1. If you have a repeat with ``cancel_on_level_end=True`` and you change the map, the repeat will be stopped, but its status remains in the old state (e.g. ``RUNNING`` if it was running before).

Ayuto:
Everything works fine, but I do have two questions:

1. It doesn't seem like we need ``_adjusted_loops`` or is just the getter missing?
2. Do we want to make ``_interval`` public or add a getter?

satoon101:
1. We could add a getter for __*_adjusted_loops*__, that would be fine with me.  If we do that, then we might also want to change __*_total_loops*__ to __*_original_loops*__ (or something like that) and add a getter for it.

2. I think a getter would be best for __*_interval*__.

Ayuto:
Alright, I will go ahead and make the changes for no. 1. @ThaPwned has asked me if we could make ``interval`` public, because he would like to change it. Of course, one could stop the repeat and restart it with a new interval, but that has a different behaviour than changing it on the fly. In that case it would still finish the current delay with the old interval and then start the next one with the new interval. So, are you fine with making it public?

satoon101:
Allowing it to be public would throw off the __*total_time*__.  To fix that, we could store each expired loop's interval in a list (or have another private attribute we use to count each interval) and use the current interval value for any remaining loops, but then comparing across loops would get much trickier.  I would much rather people do what you stated, stop the current repeat and start a new one.

Ayuto:
Oh yes, you are right. Haven't thought of that.

Issue #137, opened 2016-07-08
player_weapons_update2: weapon.clip returns 255 for weapon_knife on CS:GO


KirillMysnik:
The problem is that CS:GO uses get_property_uchar/set_property_uchar for getting/setting ammo and clips.

So [the following code in player_weapons_update2](https://github.com/Source-Python-Dev-Team/Source.Python/blob/player_weapons_update2/addons/source-python/packages/source-python/weapons/_entity/__init__.py#L29) does not work:

``` python
    def get_clip(self):
        """Return the amount of ammo in the weapon's clip."""
        return self.clip if self.clip != -1 else 0

    def set_clip(self, value):
        """Set the amount of ammo in the weapon's clip."""
        # Does the weapon have ammo?
        if self.clip != -1:

            # Set the clip amount
            self.clip = value
```

As `self.clip` always returns 255.

Edit: possible solution might be using 

``` python
return self.clip if self.clip % 256 != 255 else 0
```

KirillMysnik:
A little update on issue message: ammo is specifically ushort property in CS:GO, not uchar. That's true for grenades, as for other weapons `.primary_ammo_count` is used instead.

satoon101:
It's the same for weapon_c4 in CS:GO, as well.  In CS:S, it returns 30 for weapon_c4.  Maybe we should utilize the weapon data we store for verifying if a weapon has ammo/clip values.

```python
    def _validate_clip(self):
        """Test if the weapon has a clip."""
        if (
            self.classname in weapon_manager and
            weapon_manager[self.classname].clip is None
        ) or self._clip == -1:
            raise ValueError('Weapon does not have a clip.')

    def _validate_ammo(self):
        """Test if the weapon has a valid ammoprop and an owner."""
        if (
            self.classname in weapon_manager and
            weapon_manager[self.classname].ammoprop is None
        ) or self.ammoprop == -1:
            raise ValueError(
                'Unable to get ammoprop for {weapon}'.format(
                    weapon=self.classname
                )
            )
```

This, of course, will require us to change the default for clip to None instead of 0, which is probably the right thing to do anyway.

satoon101:
This is implemented in 6559950.  I forgot to reference this issue in the commit message.  Please test this out and post your thoughts.

invincibleqc:
I don't necessary think we should store that ourself when we can access it from their WeaponInfo instance. Testing for ammo could simply be done using:

```python
[...]

PR #160, opened 2016-10-19
Implemented has_clip/ammo & has_secondary_fire_clip/ammo to the Weapon class

KirillMysnik:
As was discussed in #137.


Ayuto:
I should have posted my opinion on this in #137, but I forgot to do so. In my opinion, I wouldn't add those methods due to EAFP:
https://docs.python.org/3/glossary.html#term-eafp

Moreover, the validation is now done twice (if you call the check method and when you actually get/set clip/ammo). That are unnecessary performance costs.

PR #169, opened 2016-12-04
CS:GO weapons update

satoon101:
* Adds item_definition_index attribute for Weapon instances for CS:GO
* Adds item_definition_index values for weapons in csgo weapon data
* Fixes maxammo on CS:GO which no longer uses ConVars
* Adds weapons to csgo weapon data that use the item_definition_index to set along with the parent classname
* Overrides created for Entity.create and Entity.find for csgo
* Adds Weapon.get_weapon_name to return the full name.  Uses the item_definition_index for CS:GO.  For other games/engines, just returns the classname


satoon101:
Currently, this uses get_weapon_name.  I'm thinking we should change it to a getter property __*weapon_name*__, but I'm open to other suggestions for the name.

Ayuto:
Yes, ``weapon_name`` would be great. I haven't tested the changes yet, but I'm afraid that ``find`` is now extremely slow, because it's utilizing ``EntityIter``. This isn't really an issue of your implementation, but rather an issue with ``__getattr__`` (and ``__setattr__``) and maybe ``EntityIter``. However, maybe we should use ``BaseEntity.get_network_property_*`` instead of accessing ``item_definition_index`` to speed things up.

satoon101:
Good point!  I tested a couple different solutions, and the most recent commit (586738c) was by far the fastest.

satoon101:
Let me know when you have had a chance to test everything, and if it's all good, we can merge this.

Ayuto:
Thanks for the update! Have you tried BaseEntityGenerator? It should be even faster than BaseEntityIter and it also provides the classname filter.

Ayuto:
I just noticed that it doesn't provide the classname filter. However, this should be a lot faster:
```python
for edict in EntityGenerator(<classname>, True):
    entity = edict.server_unknown.base_entity
    break
```

satoon101:
I tried using EntityGenerator as in your snippet, but `time.time()` was acting really funky and I couldn't get proper speed testing results.  I did find that the newest update (75c5c2d), which utilizes your earlier suggestion of BaseEntityGenerator, does significantly improve performance.  I just added the classname filter directly to the loops in Entity.find.

Ayuto:
I usually use the ``timeit`` module to measure performance. This little snippet shows that ``EntityGenerator`` is faster than ``BaseEntityGenerator``:
```python
import timeit

print(timeit.Timer(
"""
for entity in BaseEntityGenerator():
    if entity.classname != 'INVALID':
        continue
    if not entity.is_networked():
        continue
        
    break
""",
"""
from filters.entities import BaseEntityGenerator
""").timeit(ITERATIONS))

print(timeit.Timer(
"""
for edict in EntityGenerator('INVALID', True):
    entity = edict.server_unknown.base_entity
    break
""",
"""
from filters.entities import EntityGenerator
""").timeit(ITERATIONS))
```
Result:
```
4.368229347527859
0.1648294557697536
```
I used the classname ``INVALID`` to test the worst case (iterating over all entities, because no classname matches).

Ayuto:
Just added classname filter options to ``BaseEntityGenerator`` to speed things up:
```python
import timeit

ITERATIONS = 10000

print(timeit.Timer(
"""
for entity in BaseEntityGenerator():
    if entity.classname != 'INVALID':
        continue
        
    break
""",
"""
from filters.entities import BaseEntityGenerator
""").timeit(ITERATIONS))

print(timeit.Timer(
"""
for entity in BaseEntityGenerator('INVALID', True):
    break
""",
"""
from filters.entities import BaseEntityGenerator
""").timeit(ITERATIONS))
```
Result:
```
3.3641884975067313
0.13046310604846667
```
Now, there is probably no big difference between ``EntityGenerator`` and ``BaseEntityGenerator``.

satoon101:
My tests show that BaseEntityGenerator is slightly faster than EntityGenerator.  We don't really need to worry about invalid classnames in our instance, because unless we accidentally add invalid classnames to our weapon data, they will always skip both the `if` and `elif` clauses.

PR #176, opened 2017-01-02
Plugin info update

Ayuto:
- PluginInfo now has some static attributes.
- PluginInfo is now created via an ``info.ini`` file that is stored in the plugin's main directory.
- PluginInfo instances can also be retrieved for plugins that are not loaded.
- Fixed ``get_plugin_instance()`` not being able to retrieve the instance during the load process of a plugin.
- Added ``get_plugin_info()``.
- Added property ``plugins_directory``.
- Added ``plugin_exists()``.
- Added ``get_plugin_directory()``.
- Added a few attributes to ``LoadedPlugin``.
- Added the possibility to use ``__name__`` in ``get_plugin_instance`` and ``get_plugin_info``.

I think this is ready to get merged. There is only one last question that might need to be discussed: currently, a public convar is created for every plugin that is loaded unless it's explicitly disabled by setting ``public_convar`` to ``False``. Do we want this behaviour or do we want to set the default value to ``False``? I would keep the current default value (``True``).

Please also test it on your own. If there is nothing else to do, I will create a wiki page to explain the new info style a little bit more.


satoon101:
This all looks good and works well for my testing.  I would agree with leaving the default as True.

PR #178, opened 2017-01-14
Plugins changes

Ayuto:
* Untangled LoadedPlugin, SubCommandManager and PluginManager. PluginManager and LoadedPlugin now do all the logic. SubCommandManager only implements the console commands and messaging.
* Added/updated documentation
* Added OnPluginLoading/OnPluginUnloading
* Added first part of a "plugins" tutorial
* Added CommandParser.set_node_description()
* Moved CoreCommandManager.print_plugins() to SubCommandManager
* core_command is now public
* Removed LoadedPlugin.globals
* Renamed LoadedPlugin._plugin to module
* Added LoadedPlugin.unload()
* Added LoadedPlugin.reload()
* Added PluginManager.get_plugin_name()
* Added PluginManager.is_valid_plugin_name()
* Added PluginManager.get_plugin_file_path()
* Added PluginManager.plugins property
* Renamed LoadedPlugin to Plugin
* Added Plugin.sub_plugin


satoon101:
I'm not sure if this is completely done, or if you have more to do with it, so I won't merge it in.  For my testing, everything works great, and I really like the changes.

It did take me a bit to modify GunGame to get it working, since GunGame inherits from all of these classes, but I did manage it.

The only thing I would change is to move calling the plugin (un)loaded listeners calls to their own methods that can be overwritten.  It's not a huge deal, but for GunGame specifically, I would rather call and listen for events (gg_plugin_(un)loaded) than have a listener that is called for each and every plugin (un)loaded by SP or any other inheriting plugins.  I can manage this myself by adding those calls elsewhere (which I have for the time being), but it would be more convenient to just overwrite the PluginManager.notify_(un)load methods.

Ayuto:
Yes, this was meant to be in a merge-able state. After your last statement I reconsidered the listeners you mentioned, but came to the conclusion that they shouldn't be overwritten. Otherwise the listeners would have inconsistent behaviour, because some sub-plugins are firing it and some aren't. The best you can do is something like this:
```python

from listeners import OnPluginLoaded
from listeners import OnPluginUnloaded

@OnPluginLoaded
def on_plugin_loaded(plugin):
    if plugin.manager is not gg_plugin_manager:
        return
        
    # Fire gg_plugin_loaded event here...

@OnPluginUnloaded
def on_plugin_unloaded(plugin):
    if plugin.manager is not gg_plugin_manager:
        return
        
    # Fire gg_plugin_unloaded event here...
```

satoon101:
Yeah, that makes sense to not overwrite.  I'm handling it differently than that, though, as there is no need to even have that check if I fire my event right after the plugins (un)load anyway.

PR #184, opened 2017-02-09
Fixes weapon restriction

ThaPwned:
More information:
http://forums.sourcepython.com/viewtopic.php?f=20&p=9791#p9791

### DIFF
--- a/addons/source-python/packages/source-python/weapons/restrictions.py
+++ b/addons/source-python/packages/source-python/weapons/restrictions.py
@@ -190,6 +190,10 @@ class _TeamRestrictions(dict):
         raise KeyError(
             '"{0}" is neither a team number nor an alias.'.format(item))
 
+    def clear(self):
+        for x in self.values():
+            x.clear()
+
 
 class WeaponRestrictionHandler(AutoUnload):
     """Class used to handle player and team restrictions."""
@@ -203,9 +207,6 @@ class WeaponRestrictionHandler(AutoUnload):
 
     def clear(self):
         """Remove all team and player restrictions."""
-        for x in self.team_restrictions.values():
-            x.clear()
-
         self.player_restrictions.clear()
         self.team_restrictions.clear()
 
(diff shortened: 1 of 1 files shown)


satoon101:
This will have the same issue that already exists, as it is still clearing the _TeamRestrictions dictionaries.  The ```for loop``` is fine by itself as a fix, or you can instead override _TeamRestrictions.clear() to do the same ```for loop``` internally.  I think option 2 would be the better choice in case other developers want to override WeaponRestrictionHandler's clear method.

ThaPwned:
> This will have the same issue that already exists

It won't. Have been running my test server with the code for quite a while now without any exceptions but I'll update the code with your suggestion.

satoon101:
You must not have tested with a map change, then.

In [OnLevelInit](https://github.com/Source-Python-Dev-Team/Source.Python/blob/79593cbf43dfd444d9484c5daaa355e5d82ef8ea/addons/source-python/packages/source-python/weapons/restrictions.py#L397-#L400), we [clear the all of the registered handlers](https://github.com/Source-Python-Dev-Team/Source.Python/blob/79593cbf43dfd444d9484c5daaa355e5d82ef8ea/addons/source-python/packages/source-python/weapons/restrictions.py#L158-#L161).  So, that in turn [clears the team_restrictions of all items](https://github.com/Source-Python-Dev-Team/Source.Python/blob/79593cbf43dfd444d9484c5daaa355e5d82ef8ea/addons/source-python/packages/source-python/weapons/restrictions.py#L204-#L207) for each handler.  Since there are no more items in the team_restrictions, `__getitem__` [will error](https://github.com/Source-Python-Dev-Team/Source.Python/blob/79593cbf43dfd444d9484c5daaa355e5d82ef8ea/addons/source-python/packages/source-python/weapons/restrictions.py#L179-#L191) whether you pass a team number or an alias.

Thank you for the update!

ThaPwned:
I could've sworn it did... I take my statement back.
> Thank you for the update!

You're welcome.

PR #187, opened 2017-03-04
Added SetModel vfunc offsets

Predz:
Found this useful for setting the model of weapons when in the world. Hopefully should help others too :)

### DIFF
--- a/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
@@ -3,6 +3,12 @@ srv_check = False
 
 [virtual_function]
 
+    # _ZN11CBaseEntity8SetModelEPKc
+    [[set_model]]
+        offset_linux = 27
+        offset_windows = 26
+        arguments = STRING
+
     # _ZN11CBaseEntity9SetParentEPS_i
     [[set_parent]]
         offset_linux = 39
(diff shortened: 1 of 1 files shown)


addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
```diff
@@ -3,6 +3,12 @@ srv_check = False
 
 [virtual_function]
 
+    # _ZN11CBaseEntity8SetModelEPKc
+    [[set_model]]
```
Predz: Oops just noticed my naming problem 😨

satoon101:
Have you tried setting the model via the [world_model_index](https://github.com/Source-Python-Dev-Team/Source.Python/blob/313ae21031485ebeb19b8722c325b360f2168447/addons/source-python/data/source-python/entities/CBaseCombatWeapon.ini#L16) attribute?  If that works, I think it would be better to use it instead of having to rely on a virtual offset that can change whenever Valve updates the game.

```python
from engines.precache import Model

model = Model('some/model/path.mdl')
weapon = Weapon.create('weapon_flashbang')
weapon.world_model_index = model.index
```

Predz:
Yeh I agree with you and I had noticed you made that commit a while ago. However after trying to change the world model upon OnEntitySpawned, it would only work using the vfunc. So thought it was best to provide it for others.

`@OnEntitySpawned
def _change_model_world_model(entity):
	'Changes a weapons world model upon being created.'
        ## 
	if entity.classname not in worldmodel_manager:
		return
	model = worldmodel_manager[entity.classname]
	weapon = Entity(entity.index)
	weapon._set_model(model.path)`

PR #188, opened 2017-03-12
Player settings update

satoon101:
* Adds Boolean settings
  * Yes/No menu options
* Adds menu for Integer settings
  * +/- values to edit the player's current value
  * Save option to save the value
* Removes Float settings
  * If someone comes up with a viable reason to use Float settings, we can look to add this back in


satoon101:
I didn't exactly remove `*args` support, as `__new__` still allows for it.  The `__init__` method never really used the args in the past.  When calling `super().__init__`, you just have to only pass in the 3 arguments that are necessary.  And it never had `**kwargs`, unless you are talking about the `BoolSetting` class.  In that specific case, I decided to explicitly use the 3 arguments that are required for the `super` class, as `BoolSetting` doesn't need anything outside of those 3 values.

As for the named arguments, I added in the missing `super().__init__` calls and then noticed that that one was different since it used the named arguments.  I decided to make them consistent and chose not to use named arguments.  I typically do not use named arguments when the arguments use the exact same names as the variables I am passing in and it all fits on one line.

PR #201, opened 2017-06-04
Server output hook

Ayuto:
This PR introduces a server console output hook (``OnServerOutput``). You can decide whether you want to block the message from being printed or not.
Moreover, this adds a new context manager (``server_output``), which allows to intercept all messages sent during the execution of the with-statement. This allows e.g. to retrieve the output of server commands.

I have tested this with CS:S, Blade, CS:GO, L4D2 and BM:S on Windows and with CS:GO on Linux. Everything seems to work fine. But I would also like you to take a look at it before merging it into the master.


satoon101:
It mostly works for me.  The one issue I have seen is that if I try to print anything outside of any if statement, the server crashes on plugin load.  I'm certain it is because I'm causing an infinite loop and nothing can really be done about that, but we should at least add documentation around that.

Ayuto:
Yeah, I guess printing shouldn't be done at all in ``OnServerOutput`` (or at least very carefully). Though, this should only be an issue if the output redirect is active.

Issue #207, opened 2017-06-19
Add a data file updater


Ayuto:
The updater should update all Source.Python data files. There are two options to update the files:
* automatically, when Source.Python is loaded (add a setting to disable the auto-update)
* manually, by using a server command

invincibleqc:
I think updating on load is better than a command that do it on the fly. A command that update the data files, also means most of the modules would need to be reloaded at run-time (or rewritten to re-read the files to ensure they are not losing data that might be carried such as containers of live objects, etc.). Doing it on load, before any setup_functions are called (except for the config) seems to be the easiest way as all the modules could remains as is.

Ayuto:
Yeah, I don't want to implement the reloading part, so the command would simply require a restart of the server. This option could be useful for development servers, which are restarted several times.

Ayuto:
What do you think about this one?
```python
import hashlib

from zipfile import ZipFile
from urllib.request import urlopen

from paths import DATA_PATH
from paths import SP_DATA_PATH

DATA_ZIP_FILE = DATA_PATH / 'source-python.zip'
LAST_SUCCESSFUL_CHECKSUM_URL = 'http://builds.sourcepython.com/job/Source.Python%20-%20Data/lastSuccessfulBuild/artifact/checksum.txt'
LAST_SUCCESSFUL_DATA_URL = 'http://builds.sourcepython.com/job/Source.Python%20-%20Data/lastSuccessfulBuild/artifact/source-python.zip'

def get_file_checksum(file_name):
    """Return the MD5 checksum of the given file.

    :param str file_name:
        Path of the file.
    :rtype: str
    """
    with open(file_name, 'rb') as f:
        return hashlib.md5(f.read()).hexdigest()

def get_latest_data_checksum(timeout=3):
    """Return the MD5 checksum of the latest data from the build server.

    :param float timeout:
        Number of seconds that need to pass until a timeout occurs.
    :rtype: str
    """
    with urlopen(LAST_SUCCESSFUL_CHECKSUM_URL, timeout=timeout) as url:
        return url.read().decode()

def download_latest_data(timeout=3):
    """Download the latest data from the build server.

    :param float timeout:
        Number of seconds that need to pass until a timeout occurs.
    """
    with urlopen(LAST_SUCCESSFUL_DATA_URL, timeout=timeout) as url:
        data = url.read()

    with DATA_ZIP_FILE.open('wb') as f:
        f.write(data)

def unpack_data():
[...]

PR #212, opened 2017-07-04
Add data updater

Ayuto:
This PR adds the feature requested in #207.

With this updater all of Source.Python's data files are getting updated as soon as the server starts and auto updating is enabled. Auto updating is enabled by default and can be disabled via ``core_settings.ini``.

See https://github.com/Source-Python-Dev-Team/Source.Python/issues/207#issuecomment-309892000 for an explanation on how it works.


Ayuto:
@satoon101 @invincibleqc @KirillMysnik 
Did anyone of you had the chance to test that? Are you fine with me merging the changes?

KirillMysnik:
To be honest, not me :( I've only read it all just now.

invincibleqc:
I didn't test but looked at the commits when you pushed them and it sounds good to me.

PR #220, opened 2017-08-23
Update set_name offsets

njanke96:
Found in [sm](https://github.com/alliedmodders/sourcemod/blob/master/gamedata/sdktools.games/engine.csgo.txt#L188) source.


Ayuto:
Thanks you! I will be updating all the other offsets today (already started yesterday).

PR #222, opened 2017-11-03
Fix Sound.is_precached and StreamSound.is_precached performance

### LINKED ISSUE #221: Sound.is_precached and StreamSound.is_precached performance
These methods try to iterate over `stringtables.string_tables.soundprecache`, which takes a while on the Python side (~0.005sec depending on the stringtable size, easily enough to delay game frames).

jsza:
Fixes #221.

I feel like I'm missing something with this fix, but it works with some light testing.

### DIFF
--- a/addons/source-python/packages/source-python/engines/sound.py
+++ b/addons/source-python/packages/source-python/engines/sound.py
@@ -34,7 +34,7 @@ from mathlib import NULL_VECTOR
 #   Paths
 from paths import GAME_PATH
 #   Stringtables
-from stringtables import string_tables
+from stringtables import INVALID_STRING_INDEX, string_tables
 from stringtables.downloads import Downloadables
 
 
@@ -323,7 +323,7 @@ class Sound(_BaseSound):
 
         :rtype: bool
         """
-        return self.sample in string_tables.soundprecache
+        return string_tables.soundprecache[self.sample] != INVALID_STRING_INDEX
 
 
 class StreamSound(_BaseSound):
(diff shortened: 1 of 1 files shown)


jsza:
Would it perhaps be better to implement `StringTable.__contains__` on the C++ side?

invincibleqc:
Hey, thank you! I can see where it was not optimized at all with the iteration. However, an even faster way would probably be to use [engine_sound.is_sound_precached](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/engines/engines_sound_wrap.cpp#L87-L90).

jsza:
@invincibleqc Just tested on TF2 and CS:S, it looks like `engine_sound.is_sound_precached` always returns True for any string. I'll add a comment to the code warning about that.

Ayuto:
Yep, __contains__ would be the best solution. If the engine function always returns True, we should remove it.

invincibleqc:
Yes, after searching I found a [thread](https://forums.sourcepython.com/viewtopic.php?p=3401#p3401) that mentions the issue with `is_sound_precached` and is the reason why it was switched to use the string tables. Exposing a `__contains__` method on the c++ side could be a tiny bit faster but I think the difference from your proposal would be negligible. I went ahead and merged your pull request, thank you!

invincibleqc:
After thinking about it, went ahead and added a `__contains__` wrapper directly to the `StringTable` class because while this PR was fixing this particular case, the performance issue was directly in the string tables implementation.

As for `is_sound_precached`, I'm tempted to propose wrapping it on the engines affected over removing it drastically.

PR #229, opened 2017-11-26
Weapon Data: Fix CS:GO clip & maxammo values for negev, hkp2000 and tec9

backraw:
During testing of my plugin, I figured out that there is an inconsistency between the INI file and the actual values for said weapons. The recent weapon updates for CS:GO might have been the reason for that.


Ayuto:
Thank you!

PR #241, opened 2018-03-06
Core hooks

Ayuto:
This PR allows us to easily create entity hooks in C++. This can be done to speed up certain hooks.

Right now, only the PlayerRunCommand hook has been ported.


invincibleqc:
Nice! Optimizations are always welcome. After hovering the changes, I got few points:

- Any reason why `ISimpleEntityHook` accepts a `pre` and a `post` parameters over a single one accepting a `HookType_t`? Unless I missed something obvious, I don't think registering both for the same listener will ever be needed.

- Since the goal of your implementation is to optimize, this could directly read `m_nButtons` instead of `Player.buttons` (which would save quite a bit of looping behind the scene).

- Forgot a `not` into [sp_hooks.cpp](https://github.com/Source-Python-Dev-Team/Source.Python/pull/241/files#diff-5b966b1f8dba1bbf3abc8ceda47244e2R105) line 105.

- Into `ISimpleEntityHook::Initialize`, `true` is returned instead of `false` into the catch block line [93](https://github.com/Source-Python-Dev-Team/Source.Python/pull/241/files#diff-5b966b1f8dba1bbf3abc8ceda47244e2R93).

Ayuto:
Thanks for the feedback!

>I don't think registering both for the same listener will ever be needed.

That was exactly the reason why I added this. But you are right: ``HookType_t`` makes more sense.

>Into ISimpleEntityHook::Initialize, true is returned instead of false into the catch block line 93

This wasn't done by mistaken. If that part failed, it will probably also fail for all future entities as well. ANd we don't want it to fail over and over again.

PR #249, opened 2018-06-22
Make SayText, SayText2 and TextMsg messages reliable by default

jsza:
This adds `UserMessageCreator.send_reliable()` to make use of the new `RecipientFilter.reliable` attribute in https://github.com/Source-Python-Dev-Team/Source.Python/commit/cb7a8dc42d8293c75e3d0b4fe092475429139dc6 (thanks to @Ayuto for adding that!)

### DIFF
--- a/addons/source-python/packages/source-python/messages/base.py
+++ b/addons/source-python/packages/source-python/messages/base.py
@@ -42,7 +42,12 @@ class AttrDict(dict):
 
 
 class UserMessageCreator(AttrDict):
-    """Provide an easy interface to create user messages."""
+    """Provide an easy interface to create user messages.
+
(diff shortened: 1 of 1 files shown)


Ayuto:
I have never noticed that a message wasn't sent. But if you did, I would rather set ``reliable`` to ``True`` in all cases and don't let the developer decide by adding another send method.

Ayuto:
I think I would prefer this:
>I'd suggest that if we're going to go with setting reliable by default, we should do so for specific messages like SayText and SayText2

jsza:
PR updated. I also made TextMsg reliable, as it seems to have console and chat destinations.

Ayuto:
Thank you!

PR #251, opened 2018-07-11
Added CPointWorldText data

vinci6k:
Just adding some keyvalues and properties for the awesome [point_worldtext](https://developer.valvesoftware.com/wiki/Point_worldtext) entity.
The **message** keyvalue doesn't work unless it's hardcoded, not sure why.


Ayuto:
Thanks! If it doesn't work, it's not part of the DataMap. Instead it's hard-coded in ``SomeClass::KeyValue(...)``.

PR #266, opened 2018-11-29
Changed the slot values to be in line with Valve's weapon slots.

vinci6k:
Seems the values were offset by 1.

### DIFF
--- a/addons/source-python/data/source-python/weapons/constants/orangebox/cstrike/WeaponSlot.ini
+++ b/addons/source-python/data/source-python/weapons/constants/orangebox/cstrike/WeaponSlot.ini
@@ -1,7 +1,7 @@
 # ../data/source-python/weapons/constants/orangebox/cstrike/WeaponSlot.ini
 
-PRIMARY = 1
-SECONDARY = 2
-KNIFE = 3
-GRENADE = 4
-C4 = 5
+PRIMARY = 0
+SECONDARY = 1
+KNIFE = 2
+GRENADE = 3
+C4 = 4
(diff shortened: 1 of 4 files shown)


Ayuto:
The values are still correct. Why do you think they have changed?

vinci6k:
When I run this (with the current WeaponSlot.ini):
```python
# ../weapon_slot/weapon_slot.py

# Source.Python
from weapons.constants import WeaponSlot
from weapons.manager import weapon_manager
from commands.say import SayCommand


player_instances = PlayerDictionary()


@SayCommand('test')
def test_cmd(command, index, team_only=False):
    player = player_instances[index]
    weapon = player.active_weapon

    print(weapon.classname)
    print(WeaponSlot(weapon_manager.get(weapon.classname).slot))
```
I'm getting this as the output:
```
Tesla: test
weapon_m4a1
WeaponSlot.SECONDARY

Tesla: test
weapon_glock
WeaponSlot.KNIFE

Tesla: test
weapon_knife
WeaponSlot.GRENADE
```

Ayuto:
Ah, I see. However, while updating ``WeaponSlot.ini`` would fix the issue, we won't be in line anymore with Valve's weapon slots. To fix the issue and still stay in line with Valve, we would need to update this file:
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/data/source-python/weapons/cstrike.ini

This snippet retrieves Valve's weapon slots:
```python
from memory import DataType, make_object, Convention
from core import PLATFORM

player = Player(1)

Weapon_GetSlot = player.pointer.make_virtual_function(
    268 if PLATFORM == 'windows' else 269,
    Convention.THISCALL,
    [DataType.POINTER, DataType.INT],
    DataType.POINTER)
    
for x in range(5):
    ptr = Weapon_GetSlot(player, x)
    if ptr:
        weapon = make_object(Weapon, ptr)
        print(x, weapon.classname)
```

vinci6k:
I guess I was used to the keybinds for weapon switching in games, which is why I thought WeaponSlot.ini was incorrect.

I've changed WeaponSlot.ini back to what it was, and I've updated the slot values for CSS, CSGO, and HL2DM.

Ayuto:
Thanks!

PR #268, opened 2018-12-20
Added new weapons to CSGO data.

vinci6k:
Added all the missing weapons, but there are two things I couldn't find: the slot for weapon_tablet (it's bound to the same key as the scoreboard) and the new slot for weapon_healthshot (used to be on the same slot as the c4, but it's been moved to slot12 (console command), which is bound to X by default).


vinci6k:
Scratch that, using [Weapon_GetSlot](https://github.com/Source-Python-Dev-Team/Source.Python/pull/266#issuecomment-442994762) I was able to get the actual slot values for both of these weapons.

Ayuto:
Thank you!

PR #272, opened 2019-02-08
Added from_inthandle() to the Entity and EntityDictionary classes.

vinci6k:
Similarly to from_userid() that the [Player](https://github.com/Source-Python-Dev-Team/Source.Python/blob/972547be070e74a3f785284cefef53d98fab83c8/addons/source-python/packages/source-python/players/_base.py#L91) and [PlayerDictionary](https://github.com/Source-Python-Dev-Team/Source.Python/blob/972547be070e74a3f785284cefef53d98fab83c8/addons/source-python/packages/source-python/players/dictionary.py#L33) have, I think this can be very useful for plugins that work with active weapons a lot.

```python
from events import Event
from weapons.dictionary import WeaponDictionary


player_instances = PlayerDictionary()
weapon_instances = WeaponDictionary()


@Event('weapon_fire')
def weapon_fire(event):
    player = player_instances.from_userid(event['userid'])
    weapon = weapon_instances.from_inthandle(player.active_weapon_handle)
```

### DIFF
--- a/addons/source-python/packages/source-python/weapons/_base.py
+++ b/addons/source-python/packages/source-python/weapons/_base.py
@@ -8,6 +8,7 @@
 # Source.Python Imports
 #   Entities
 from entities.entity import Entity
+from entities.helpers import index_from_inthandle
 #   Weapons
 from weapons.manager import weapon_manager
 
@@ -25,6 +26,15 @@ __all__ = ('Weapon',
 # =============================================================================
 class Weapon(Entity):
     """Allows easy usage of the weapon's attributes."""
+
+    @classmethod
+    def from_inthandle(cls, inthandle):
+        """Create an instance from an inthandle.
+        :param int inthandle:
+            The inthandle.
+        :rtype: Weapon
+        """
+        return cls(index_from_inthandle(inthandle))
 
     def _validate_clip(self):
         """Test if the weapon has a clip."""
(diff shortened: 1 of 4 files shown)


addons/source-python/packages/source-python/weapons/_base.py
```diff
@@ -26,6 +27,15 @@
...
 
+    @classmethod
+    def from_inthandle(cls, inthandle):
+        """Create an instance from an inthandle.
+        :param int inthandle:
+            The inthandle.
+        :rtype: Weapon
+        """
+        return cls(index_from_inthandle(inthandle))
+
```
Ayuto: I think it would make more sense if that would be added to the Entity class.

addons/source-python/packages/source-python/weapons/dictionary.py
```diff
@@ -28,3 +29,11 @@ class WeaponDictionary(EntityDictionary):
...
         """Initialize the dictionary."""
         super().__init__(factory, *args, **kwargs)
+
+    def from_inthandle(self, inthandle):
+        """Get a weapon instance from an inthandle.
+        
+        :param int inthandle: The inthandle.
+        :rtype: Weapon
+        """
+        return self[index_from_inthandle(inthandle)]
```
Ayuto: Same for this one. It should be added to EntityDictionary.

vinci6k:
Oh, right! I completely forgot about the other inthandles (m_hGroundEntity, m_hOwnerEntity, m_hOwner).
I've moved them to the Entity and EntityDictionary classes.

Ayuto:
Thanks!

It's not because of these three attributes. Even if I couldn't provide a single scenario I would add it to those two base classes. Simply because of the possible existance of such scenarios.

PR #273, opened 2019-02-25
Added CBaseCombatCharacter::OnTakeDamage_Alive data.

vinci6k:
When working with armor (CSS, CSGO) or critical hits (TF2), [CBaseEntity::OnTakeDamage](https://github.com/alliedmodders/hl2sdk/blob/0ef5d3d482157bc0bb3aafd37c08961373f87bfd/game/server/baseentity.cpp#L1376) gives incorrect values. I found out that [CBaseCombatCharacter::OnTakeDamage_Alive](https://github.com/alliedmodders/hl2sdk/blob/0ef5d3d482157bc0bb3aafd37c08961373f87bfd/game/server/basecombatcharacter.cpp#L2463) gives the correct values, so I've added the data for all supported games.

```python
# ../damage_comparison/damage_comparison.py

# Source.Python
from entities import TakeDamageInfo
from entities.entity import BaseEntity


@EntityPreHook(EntityCondition.is_player, 'on_take_damage')
def on_take_damage_pre(stack_data):
    base_entity = BaseEntity._obj(stack_data[0])
    info = TakeDamageInfo._obj(stack_data[1])

    print(f'otd:   player({base_entity.index}) took {info.damage} damage')


@EntityPreHook(EntityCondition.is_player, 'on_take_damage_alive')
def on_take_damage_alive_pre(stack_data):
    base_entity = BaseEntity._obj(stack_data[0])
    info = TakeDamageInfo._obj(stack_data[1])

    print(f'otd_a: player({base_entity.index}) took {info.damage} damage')
```
Output (from CSGO):
```
otd:   player(5) took 28.294065475463867 damage
otd_a: player(5) took 13.0 damage
```
I wasn't sure where to add the data for Blade Symphony, but it does have the same offset as Left 4 Dead 2, are these two games on the same Source engine branch?

### DIFF
--- a/addons/source-python/data/source-python/entities/bms/CBaseCombatCharacter.ini
+++ b/addons/source-python/data/source-python/entities/bms/CBaseCombatCharacter.ini
@@ -1,4 +1,11 @@
 [virtual_function]
+    
+    # _ZN20CBaseCombatCharacter18OnTakeDamage_AliveERK15CTakeDamageInfo
+    [[on_take_damage_alive]]
+        offset_linux = 273
+        offset_windows = 272
+        arguments = POINTER
+        return_type = INT
 
     # _ZN20CBaseCombatCharacter13Weapon_SwitchEP17CBaseCombatWeaponi
     [[weapon_switch]]
--- a/addons/source-python/data/source-python/entities/csgo/CBaseCombatCharacter.ini
+++ b/addons/source-python/data/source-python/entities/csgo/CBaseCombatCharacter.ini
@@ -2,6 +2,13 @@ srv_check = False
 
 
 [virtual_function]
+    
+    # _ZN20CBaseCombatCharacter18OnTakeDamage_AliveERK15CTakeDamageInfo
+    [[on_take_damage_alive]]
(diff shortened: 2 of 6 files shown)


invincibleqc:
Thanks!

> I wasn't sure where to add the data for Blade Symphony, but it does have the same offset as Left 4 Dead 2, are these two games on the same Source engine branch?

The data is parsed using the [GameConfigObj](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/core/__init__.py#L143-L161) class which merge and overwrite the data using the `shared → engine → game` tree. Since Blade Symphony is defining `SOURCE_ENGINE` to `"blade"` in the [blade.cmake file](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/makefiles/branch/blade.cmake#L9), the data for the `CBaseCombatCharacter` class for that engine would go into `../blade/CBaseCombatCharacter.ini`. We don't have any data specific to that engine, and I honestly don't think anyone even play that game anymore. Mainly explained for future reference to how it internally works.

PR #276, opened 2019-04-19
Player class updates

Ayuto:
Just like the [Entity changes](https://github.com/Source-Python-Dev-Team/Source.Python/pull/235) last year, but now for the Player class.

This PR removes the properties from the data files that have been added to the BaseEntity class statically, because we don't need them anymore.

The PR also moves all player properties to a class called PlayerMixin. The PlayerMixin class is mixed into the class ``players._base.Player``. This is done to gain major speed improvements (85 - 125 times faster depending on the property). Moreover, these properties are now documented on the wiki.

I have also moved the game specific properties to PlayerMixin. If you access one of those properties in a different game, you will get the following error:
``ValueError: Unable to find property 'm_Shared.m_iDesiredPlayerClass'.``

We could think about catching this exception and re-raising a NotImplementedError or something like that. What's your opinion on this?

Outlook:
- Return the proper entity class for some attributes (e. g. ``observer_target``) instead of returning an index or handle.
- Move more player functions/properties to the mixin class

PS: You need an instance of the Player class to take advantage of the speed improvements. If you use the Entity class, it will still access the data files. That's something I can live with for now.

### DIFF
--- a/src/core/modules/players/players_entity.cpp
+++ b/src/core/modules/players/players_entity.cpp
@@ -38,7 +38,7 @@ boost::shared_ptr<PlayerWrapper> PlayerWrapper::__init__(unsigned int uiEntityIn
 {
 	CBaseEntityWrapper* pEntity = (CBaseEntityWrapper*) ExcBaseEntityFromIndex(uiEntityIndex);
 	if (!pEntity->IsPlayer())
-		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Index '%d' is not a valid player.");
+		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Index '%d' is not a valid player.", uiEntityIndex);
 
 	return PlayerWrapper::wrap(pEntity->GetBaseEntity());
 }
--- a/src/core/modules/players/players_entity.cpp
+++ b/src/core/modules/players/players_entity.cpp
@@ -32,294 +32,294 @@
 
 
 // ============================================================================
-// >> PlayerWrapper
+// >> PlayerMixin
 // ============================================================================
-boost::shared_ptr<PlayerWrapper> PlayerWrapper::__init__(unsigned int uiEntityIndex)
+boost::shared_ptr<PlayerMixin> PlayerMixin::__init__(unsigned int uiEntityIndex)
 {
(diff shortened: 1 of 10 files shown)


src/core/modules/players/players_entity.cpp
```diff
@@ -0,0 +1,702 @@
...
+// CBasePlayer
+float PlayerWrapper::GetSpeed()
+{
+	static int offset = FindNetworkPropertyOffset("localdata.m_flLaggedMovementValue");
+	return GetNetworkPropertyByOffset<float>(offset);
+}
+
+void PlayerWrapper::SetSpeed(float value)
+{
+	static int offset = FindNetworkPropertyOffset("m_Local.m_bDucked");
```
satoon101: This should be `localdata.m_flLaggedMovementValue`.  However, when testing this branch with GunGame locally, it errored saying:

```python
[SP] Caught an Exception:
  File "..\addons\source-python\packages\source-python\events\listener.py", line 92, in fire_game_event
    callback(game_event)
  File "..\addons\source-python\plugins\gungame\plugins\included\gg_multi_level\gg_multi_level.py", line 156, in _player_level_up
    multi_level_manager.give_multi_level(player.userid)
  File "..\addons\source-python\plugins\gungame\plugins\included\gg_multi_level\gg_multi_level.py", line 128, in give_multi_level
    self[userid] = _MultiLevelPlayer.from_userid(userid)
  File "..\addons\source-python\packages\source-python\players\_base.py", line 99, in from_userid
    return cls(index_from_userid(userid))
  File "..\addons\source-python\plugins\gungame\plugins\included\gg_multi_level\gg_multi_level.py", line 59, in __init__
    self.speed = speed.get_int() / 100
  File "..\addons\source-python\packages\source-python\entities\_base.py", line 125, in __setattr__
    object.__setattr__(self, attr, value)

ValueError: Unable to find property 'm_Local.m_bDucked'.
```

So, I'm not sure if ducking is in a working state.  This was on CS:S.

Ayuto: Good catch! I think it's still a valid property, but it's a datamap property, not a networked property. I will fix this when I get home on Friday.

Ayuto: Fixed it via Github Web.

PR #277, opened 2019-05-31
Added engine_import implementation.

jordanbriere:
This PR addresses the documentation issues that were discussed into the following thread: https://forums.sourcepython.com/viewtopic.php?p=10831#p10831


Ayuto:
Looks good to me! Though, it might be worth mentioning what directory structure ``engine_import`` expects.

addons/source-python/packages/source-python/core/__init__.py
```diff
@@ -330,4 +349,79 @@ def check_info_output(output):
...
+                continue
+            if (attr == '__doc__' and
+                    getattr(caller, '__doc__', None) is not None):
+                continue
+            if attr == '__all__':
+                if hasattr(caller, '__all__'):
+                    obj = tuple(sorted(set(obj + getattr(caller, '__all__'))))
+            elif skip_privates and attr.startswith('_'):
+                continue
+            elif isclass(obj):
```
satoon101: This line raises NotImplementedError on DOD:S and TF2 servers.  I originally thought this was because we don't have any weapon data for those 2 games, but when I tested Garrysmod, it worked fine.

```python

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\__init__.py", line 93, in load
    setup_sp_command()

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\__init__.py", line 329, in setup_sp_command
    from core.command import auth, docs, dump, plugin

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\core\command\__init__.py", line 20, in <module>
    from commands.typed import TypedServerCommand

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\commands\typed.py", line 30, in <module>
    from filters.players import parse_filter

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\filters\players.py", line 26, in <module>

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\players\entity.py", line 1021, in <module>
    engine_import()

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\core\__init__.py", line 390, in engine_import
    elif isclass(obj):

  File "E:\Servers\tf\tf\addons\source-python\Python3\inspect.py", line 78, in isclass
    return isinstance(object, type)

  File "E:\Servers\tf\tf\addons\source-python\packages\source-python\weapons\default.py", line 29, in __getattribute__
    'No support for game "{0}"'.format(GAME_NAME))

NotImplementedError: No support for game "tf"
```

If I wrap that isclass in a try/except, it works fine, but I'm not sure if there are any negative consequences of doing so.

addons/source-python/packages/source-python/weapons/entity.py
```diff
@@ -5,10 +5,14 @@
...
-from core import GAME_NAME
-from core import SOURCE_ENGINE
-from paths import SP_PACKAGES_PATH
+# Source.Python Imports
+#   Core
+from core import engine_import
+#   Entities
+from entities.entity import Entity
+#   Weapons
+from _weapons._entity import WeaponMixin
```
satoon101: When building the documentation, the _weapons module gives an ModuleNotFoundError:

```python
E:\Servers\cstrike\cstrike\addons\source-python\docs\source-python\source\developing\modules\weapons.entity.rst:9: WARNING: autodoc: failed to import class 'WeaponMixin' from module '_weapons._entity'; the following exception was raised:
  File "E:\Servers\cstrike\cstrike\addons\source-python\packages\site-packages\sphinx\ext\autodoc.py", line 518, in import_object
    __import__(self.modname)
ModuleNotFoundError: No module named '_weapons'
```

addons/source-python/packages/source-python/core/__init__.py
jordanbriere: This happens because `isclass` is internally calling `isinstance` which attempts to get the `__class__` attribute of the object but since `NoWeaponManager` is raising into `__getattribute__`, it fails. The same would occurs with:
```python
from weapons.default import NoWeaponManager
mgr = NoWeaponManager()
mgr.__class__
```
I don't believe we should wrap the exception because it is working as intended. That class is designed to raise in such scenario and I believe it should rather raise into `__getattr__` instead so that it only fails when attempting to get attributes that are not contained in the `__dict__` instance of the class itself.

addons/source-python/packages/source-python/weapons/entity.py
jordanbriere: Seems like you are running an old build that date before the commit that introduced the _weapons._entity module: https://github.com/Source-Python-Dev-Team/Source.Python/commit/d9b7b6b7cc5431650a7db1b2a47f0d916f04973d
If I correct that `__getattribute__` to `__getattr__` that was pointed above, the doc is building fine for me on TF2.

PR #286, opened 2019-11-21
Fix #285, thanks to AlliedModders/SourceMod

### LINKED ISSUE #285: [CSGO] Player objects don't works after update of 2019-11-18
I noticed that all bots on my server no longer have weapons.

for example;

```
player.give_named_item('weapon_elite') # not working
print(player.name) # UnicodeEncodeError: 'ascii' codec can't encode character '\u2606' in position 2: ordinal not in range(128)
```

The cause of the problem looks like yesterday's CS:GO update. And it looks like it points to the wrong address. There may be other similar problems.
Could you fix it?

CookStar:
Referenced from alliedmodders/sourcemod@bef8562de58cc25a52266cd51f92c74d62f26db1

Not tested on Windows.


jordanbriere:
Thank you! There is quite a bit more data to update for players on CS:GO but I will push that shortly.

PR #292, opened 2019-11-28
Added cached properties implementation.

jordanbriere:
This branch adds support for cached properties on both; Python and C++ sides. Example:

```python
from random import randint
from core.cache import cached_property

class Test:
    @cached_property(kwargs=dict(range=(0, 1000)))
    def test(self, range):
        return randint(*range)

    @test.setter
    def set_test(self, value, range):
        return int(value / 2)

test = Test()

# Compute and cache the value for the first time
i = test.test

# The first computed value was cached, so it should always be the same
assert i is test.test
assert i is test.test
assert i is test.test
assert i is test.test

# Deleting the property is invalidating the cache
del test.test
assert i is not test.test

# The cache will be updated to 5, because our setter computes value / 2
test.test = 10
assert test.test is 5

# The new value should be 1, because we updated our userdata
Test.test['range'] = (1, 1)
del test.test
assert test.test is 1
```
Summary of changes compared to master:

```
Added cached properties implementation.
Entity's dynamic functions, inputs, outputs, and various properties are now cached.
Improved performance of Entity.<get/set>_property_<type> methods.
Fixed BaseEntity.<get/set>_network_property_<type> from reading/writing from a NULL pointer.
Fixed excluded properties/collapsible tables not being skipped.
Fixed memory leaks related to Function's convention.
Added segfault handler to Binary.find_pointer.
Fixed some exceptions being silenced under certain conditions within the load/unload process.
Added BaseEntity.entity_flags.
Fixed an issue when entities instances were cached again after we invalidated the cache.
Fixed EntityDictionary potentially caching soon-to-be removed entities.
Some various improvements to some redundant logics.
Entity.__hash__ now hashes its inthandle rather than its index.
Fixed the invalidation of the internal entity cache before all entity deletion listeners were called.
Fixed a KeyError being silenced in EntityDictionary.on_automatically_removed overrides.
Slightly improved the retrieval speed of the cached entity instances.
Improved performance of memory tools (approximately 2.5 times faster).
Fixed entity delays/repeats not being cancelled if they were registered from an OnEntityDeleted listener called after the internal callback.
Player.language property is now only cached for games it was already cached.
Optimized non-cached get_client_language by using the internal Player cache instead of PlayerInfo.
Optimized various python calls from c++.
Optimized pointer extraction.
Fixed memory leaks caused by Sound/StreamSound instances never unloaded when instantiated from Entity.emit_sound.
Fixed memory leaks into engine_sound.emit_sound caused by the sounds not being emitted through the reliable channel.
Improved performance of Entity.is_in_solid's default behaviour (approx. 19 times faster).
Moved Entity.<get/set>_property_<type> methods to BaseEntity (approx. ~18 times faster).
Added BaseEntity.<get/set>_<datamap/network>_property_edict.
Added BaseEntity.<get/set>_property_<double/long/long_long/string_array/ulong/ulong_long> methods.
Improved performance of TraceFilterSimple.should_hit_entity (approx. 42 times faster).
Improved performance of Entity.create, find and find_or_create by moving them to BaseEntity directly.
Added an overload to memory.make_object to avoid redundant extraction.
Added BaseEntity.is_marked_for_deletion (approx. 13 times faster than the manual check).
Moved InputFunction to c++ (~21x faster).
Added cached_result decorator.
Improved some Entity's methods by avoiding repeated attribute retrievals.
```

<details>
 <summary>Test code:</summary>

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)

    pl.index
    pl.server_classes
    pl.properties
    pl.inputs
    pl.outputs
    pl.keyvalues

    pl.server_class
    pl.datamap
    pl.factory
    pl.edict
    pl.index
    pl.pointer
    pl.inthandle
    pl.physics_object
    pl._size

    pl.raw_steamid
    pl.playerinfo
    pl.userid
    pl.steamid
    pl.client
    pl.base_client
    pl.uniqueid
    pl.address
    pl.is_hltv
    pl.is_bot

    pl.on_take_damage
    pl.run_command
    pl.drop_weapon
    pl.give_named_item
    pl.add_account
    pl.autobuy
    pl.buy_internal
    pl.rebuy
    pl.deafen
    pl.teleport
    pl.switch_team

    for input in pl.inputs:
        pl.get_input(input)

    for output in pl.outputs:
        pl.get_output(output)

    pl.set_property_int('m_hUseEntity', pl.get_property_int('m_hUseEntity'))
    pl.set_property_int('m_iHealth', pl.get_property_int('m_iHealth'))
    pl.set_property_uchar('m_lifeState', pl.get_property_uchar('m_lifeState'))
    pl.set_property_short('m_iBonusProgress', pl.get_property_short('m_iBonusProgress'))
    pl.set_property_char('m_iBonusChallenge', pl.get_property_char('m_iBonusChallenge'))
    pl.set_property_float('m_flMaxspeed', pl.get_property_float('m_flMaxspeed'))
    pl.set_property_ushort('m_fFlags', pl.get_property_ushort('m_fFlags'))
    pl.set_property_uchar('m_iObserverMode', pl.get_property_uchar('m_iObserverMode'))
    pl.set_property_int('m_hObserverTarget', pl.get_property_int('m_hObserverTarget'))
    pl.set_property_uchar('m_iFOV', pl.get_property_uchar('m_iFOV'))

print(time() - t)
```
</details>

Results:

```
Master: 1184.23854987357252
This branch: 47.69846749305723 (~25x faster)
```

Windows test builds (https://github.com/Source-Python-Dev-Team/Source.Python/pull/292/commits/f730a62986e2b904e9cbacca772be4344af2eb76): https://drive.google.com/open?id=1Zr2eDirFmpxn3XB-4Wd5i-qYT8SbKPWQ


jordanbriere:
Well, that last one is more than welcome... 😮 

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)

    pl.on_take_damage
    pl.run_command
    pl.drop_weapon
    pl.give_named_item
    pl.add_account
    pl.autobuy
    pl.buy_internal
    pl.rebuy
    pl.deafen
    pl.teleport
    pl.switch_team

print(time() - t)
```

Results:

```
Cached: 3.031463384628296
Not cached: 307.1604826450348
```

Tested the following code on this branch versus master:

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)

    pl.on_take_damage
    pl.run_command
    pl.drop_weapon
    pl.give_named_item
    pl.add_account
    pl.autobuy
    pl.buy_internal
    pl.rebuy
    pl.deafen
    pl.teleport
    pl.switch_team

    pl.index
    pl.server_classes
    pl.properties
    pl.inputs
    pl.outputs
    pl.keyvalues

    pl.server_class
    pl.datamap
    pl.factory
    pl.edict
    pl.index
    pl.pointer
    pl.inthandle
    pl.physics_object
    pl._size

    pl.raw_steamid
    pl.playerinfo
    pl.userid
    pl.steamid
    pl.client
    pl.base_client
    pl.uniqueid
    pl.address
    pl.is_hltv
    pl.is_bot

    try:
        pl.invalid_attribute
    except:
        pass


print(time() - t)
```

Results:

```python
This branch: 20.118335723876953
Master: 350.64219880104065
```

jordanbriere:
The performance difference of the last commit can essentially be seen when iterating over the entity instances (e.g. within `__getattr__`):

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)
    for instance in pl.instances:
        pass

print(time() - t)
```
```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)
    for server_class in pl.server_classes:
        make_object(server_class, pl.pointer)

print(time() - t)
```
Results:
```
This branch: 1.0929217338562012
Master: 35.05029082298279
```

invisiblesoldiers:
I did some performance tests a week ago comparing with SourceMod and noticed that getting any entity property in 100x slower on SourcePython. Maybe after your cached properties commits it is 50x but the situation is still bad. And another test with trace rays, it was just calling C++ functions, result isn't positive and if I remember correctly SourcePython in 20x slower than SourceMod in the case. So why so?

invisiblesoldiers:
I know that cycling mind on optimizations is bad but performance is important when you type scripts for entirely game engine because need process all code less than server tick interval

jordanbriere:
Comparing SourceMod with Source.Python is like comparing a race car with a family vehicle. The former was built to be fast, while the other was built to be comfortable. Both will take you from A to B, but the later will be a smoother ride than the former. What I mean by that is, that a beginner will have less trouble starting with Source.Python than SourceMod.

SourcePawn is a compiled language, while Python is interpreted. You can find a good comparison in the following article: [Compiled Versus Interpreted Languages](https://guide.freecodecamp.org/computer-science/compiled-versus-interpreted-languages/)

A short summary could be: the interpreted language will always be slower to run, because most of its work is done at run-time while the compiled language do it at compile time.

It comes with advantages and disadvantages. The main advantage of embedding Python, is that you do not have to re-invent the wheel. Without mentioning all the built-in modules and [community libraries](https://pypi.org/) it has to offers. On the down side; it comes with a performance cost.

It is also a great learning curve for hobbyists to get into a programming career. I know a lot from the Source.Python and EventScripts communities that became professional programmers which all started from scripting random stuff for their favourite games. I mean, and I say that with all the respect in the world, learning SourcePawn is probably not something you will use outside of modding source games. While learning Python, could leads you to web development, machine learning, mobile apps, etc, etc. Python is a very powerful and versatile language while SourcePawn is, while can teaches you the logical aspect and mindset of programming, more of a niche language that is; modding source games. At least, that is my personal opinion.

So yeah, if you are looking for performance for modding your source games; SourceMod is your choice. If you are looking to learn and mess around with stuff, and perhaps diverge into other programming branches later on, then Source.Python is probably your best bet. Of course, as stated above, both can teaches you invaluable lessons, but I consider one have paths ahead while the other doesn't. But of course, it depends of what your goals are, but me personally, I stuck with EventScripts at the time for that specific reason and I remember we had a very constructive chat on the eventscripts-fr IRC channel about it (/salute Nico, hope you are doing well whatever you are doing these days!). And I personally don't regret it, as I've been using Python for multiple unrelated purposes ever since.

> I did some performance tests a week ago comparing with SourceMod and noticed that getting any entity property in 100x slower on SourcePython.

I believe it depends of your code. In fact, I don't believe I know for a fact it does. One of the problem with Source.Python, is that from the start we opted to design it to be extremely dynamic and as close as possible to the engine. Which led to duplicate tools, while some being much slower than the other making it so that basic stuff you are doing might not necessarily be the best when it comes to performance. Here is a good example:

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)
    pl.get_property_ushort('m_iAccount')

print(time() - t)
```
Result:
```
10.130887269973755 (on master, on this branch I now get 6.767904376983643)
```
It is slow, because that method is, well, first of all written in pure Python, but mainly because it tries to determine what you are requesting from it. So it successively looks the entity's networked properties, datamaps, etc. in order to find a match. But if you are more specific, for example you directly ask for a network property, you are going to get much better performance because you point your request in a specific direction rather than letting it orienting itself. For example:

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)
    pl.get_network_property_ushort('m_iAccount')

print(time() - t)
```
Result:
```
0.654240608215332
```
It also helps that the later was implemented in c++, but just to give an example showing that the way you write your code and which tools you are using can have a major impact on its performance.

Just as an addition, you could also use the alias of this property: `pl.cash`, but it would be slightly slower than the direct call because Python would first have to resolve the attribute before forwarding to its descriptor. Just another example of; convenience VS performance.

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)
    pl.cash

print(time() - t)
```
Result:
```
0.8756575584411621
```

Another thing that have major impact on the performance is probably the code being "over-sanitized". Not sure if that is an existing expression, but basically what I mean is the data being internally validated multiple times. The following commit is probably a good example of it: https://github.com/Source-Python-Dev-Team/Source.Python/commit/ec3f4247f6563b2932e5a37ee60b06dcea219f9a. Another good example would be `hasattr`. Take the following example:

```python
class Class:
    something = 'something'
    something_else = 'something else'
cls = Class()

t = time()
for i in range(100000000):
    if hasattr(cls, 'something'):
        something = getattr(cls, 'something')

print(time() - t)
```
Result:
```
28.30333185195923
```
If you are looking for performance, then this is not a good implementation. The code above is pretty much the same logic as writting:
```python
for i in range(100000000):
    try:
        something = getattr(cls, 'something')
        result = True
    except AttributeError:
        result = False
    if result:
        something = getattr(cls, 'something')
```

You basically tell Python:
> Hey, go look into data if there is something, if there is, come back to let me know then go look for it again and bring it back.

Of course, it works. But if you are looking for performance then it is bad practice as you are technically asking Python to retrieve the value twice. What you should be telling Python instead is:

> Hey, go look into data if there is something, if there is, bring it back or let me know if there was a problem.

```python
t = time()
for i in range(100000000):
    try:
        something = getattr(cls, 'something')
    except AttributeError:
        pass

print(time() - t)
```
Result:
```
15.540138721466064
```
Two birds one stone; you got your value at the same time of validating it so you greatly improved the performance. This just shows that everything has cost, and you should always try to find the fastest route if performance is an issue.

> Maybe after your cached properties commits it is 50x but the situation is still bad.

I will have to agree on that one. I knew interacting with entities was somewhat slow, but I never expected to get these results. And I can only imagine the results I would have gotten prior to Ayuto moving most of the aliases on the c++ side.

Either way, this branch is indeed trying to remedy and improve the situation, but there is always a limit of what can be done while keeping backward compatibility. My tests shows interacting with entities has an average of 1600% performance boost with this branch.

> And another test with trace rays, it was just calling C++ functions, result isn't positive

Please post on the forums, or create an issue with your code and we can investigate and see if improvements can be done either directly in your code or in Source.Python itself.

Sorry for the book.

jordanbriere:
Latest commit is caching inputs and outputs instances.

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)

    for input in pl.inputs:
        pl.get_input(input)

    for output in pl.outputs:
        pl.get_output(output)

print(time() - t)
```
Results:
```
Cached: 34.75408148765564
Master: 572.1209080219269
```

jordanbriere:
Latest commit improves the performance of `Entity.get_property_<type>` methods:

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)

    pl.set_property_int('m_hUseEntity', pl.get_property_int('m_hUseEntity'))
    pl.set_property_int('m_iHealth', pl.get_property_int('m_iHealth'))
    pl.set_property_uchar('m_lifeState', pl.get_property_uchar('m_lifeState'))
    pl.set_property_short('m_iBonusProgress', pl.get_property_short('m_iBonusProgress'))
    pl.set_property_char('m_iBonusChallenge', pl.get_property_char('m_iBonusChallenge'))
    pl.set_property_float('m_flMaxspeed', pl.get_property_float('m_flMaxspeed'))
    pl.set_property_ushort('m_fFlags', pl.get_property_ushort('m_fFlags'))
    pl.set_property_uchar('m_iObserverMode', pl.get_property_uchar('m_iObserverMode'))
    pl.set_property_int('m_hObserverTarget', pl.get_property_int('m_hObserverTarget'))
    pl.set_property_uchar('m_iFOV', pl.get_property_uchar('m_iFOV'))

print(time() - t)
```
Results:
```
This branch: 14.038818120956421
Master: 211.55949449539185
```
They could probably be improved a bit more by being written on the c++ side but the improvement is good enough for now.

That commit also fixes issues with the `BaseEntity.<get/set>_network_property_<type>` methods. For example, setting `overlay_vars` for a player would crash, because it is excluded for the player class, and is a collapsible table for `CBaseAnimatingOverlay` causing us to write the value at a NULL pointer.

Also, `CBaseEntityWrapper::FindNetworkPropertyOffset` now forwards to `CBaseEntityWrapper::FindDatamapPropertyOffset` for RecvTables/Arrays because it would also cause us to read/write a NULL pointer as they needs to be processed by their proxy functions. Ideally we would want to proxy them, but their datamap contains their raw offsets we can use with the current implementation so that is probably a good enough workaround for now.

jordanbriere:
Latest commit fixes some leaks I noticed while timing the following code:

```python
from time import time

t = time()
for i in range(1000000):
    pl = Player(1)

    pl.index
    pl.server_classes
    pl.properties
    pl.inputs
    pl.outputs
    pl.keyvalues

    pl.server_class
    pl.datamap
    pl.factory
    pl.edict
    pl.index
    pl.pointer
    pl.inthandle
    pl.physics_object
    pl._size

    pl.raw_steamid
    pl.playerinfo
    pl.userid
    pl.steamid
    pl.client
    pl.base_client
    pl.uniqueid
    pl.address
    pl.is_hltv
    pl.is_bot

    pl.on_take_damage
    pl.run_command
    pl.drop_weapon
    pl.give_named_item
    pl.add_account
    pl.autobuy
    pl.buy_internal
    pl.rebuy
    pl.deafen
    pl.teleport
    pl.switch_team

    for input in pl.inputs:
        pl.get_input(input)

    for output in pl.outputs:
        pl.get_output(output)

    pl.set_property_int('m_hUseEntity', pl.get_property_int('m_hUseEntity'))
    pl.set_property_int('m_iHealth', pl.get_property_int('m_iHealth'))
    pl.set_property_uchar('m_lifeState', pl.get_property_uchar('m_lifeState'))
    pl.set_property_short('m_iBonusProgress', pl.get_property_short('m_iBonusProgress'))
    pl.set_property_char('m_iBonusChallenge', pl.get_property_char('m_iBonusChallenge'))
    pl.set_property_float('m_flMaxspeed', pl.get_property_float('m_flMaxspeed'))
    pl.set_property_ushort('m_fFlags', pl.get_property_ushort('m_fFlags'))
    pl.set_property_uchar('m_iObserverMode', pl.get_property_uchar('m_iObserverMode'))
    pl.set_property_int('m_hObserverTarget', pl.get_property_int('m_hObserverTarget'))
    pl.set_property_uchar('m_iFOV', pl.get_property_uchar('m_iFOV'))

print(time() - t)
```

The memory usage was increasing a lot and never released. Either way, here are the time results:

```
This branch: 67.23700252525907
Master: 1184.23854987357252
```

Showing that accessing entities attributes is approximately 17 times faster than it was! 🎉

jordanbriere:
With the minor improvements made into last commit the new time I'm getting running [this code](https://github.com/Source-Python-Dev-Team/Source.Python/pull/292#issuecomment-560172986) on this branch is: `54.06634736061096`

jordanbriere:
Man. That last one gave me headaches. When testing, I noticed that if I was using `caching=False` the memory just kept growing and growing. I first thought it was caused by the back reference (which is something I will have to fix, as it renders subclassing `CachedProperty` useless as it would just returns a new proxy object of that class rather than the subclass, etc.) but it was caused by a circular reference preventing the instance itself to be freed. Pretty much like:

```python
from random import randint

class Class:
    def __init__(self):
        self.self = self

        # Generate lot of data so we can see the impact on memory.
        self.data = [randint(0, i) for i in range(100000)]

for i in range(1000000000):
    c = Class()
```

Since `c` internally stores itself, it cause a circular reference meaning it might or might not be freed. I say might or might not because Python might detect those at a some point through its traverse mechanism but there is no guarantee. But yeah, same concept as caching an object that refers to the instance it is cached into.

I also believe this was the same issue that led me to remove `CachedGenerator` into https://github.com/Source-Python-Dev-Team/Source.Python/pull/292/commits/afde4a1aa14246b2b11843dc94bb12328fce600e where I thought the entire frame was copied and never freed but I think that since the frame itself was referring to the scope which was indirectly referring to the class that our `CachedProperty` was bound to referred by the instance that the generator was cached in we were ending with a circular referrence. Or something along these lines.

Anyways, I've been playing a bit of [GunGame](https://github.com/GunGame-Dev-Team/GunGame-SP) (which does a lot of interactions with entities, players, weapons, etc.) with bots and everything is working as it used to. The average time I'm getting as of now is `52.405569553375244` meaning entity interactions is up to 22 times faster than before!

jordanbriere:
Interestingly enough, fixing the back reference reduced the average time to `50.42988157272339`! I guess because Boost no longer have to instantiate a new proxy wrapping the address to return. Anyways, I'm quite satisfied to where this branch is currently sitting at, so unless someone find a bug I didn't notice I think it is in a good state to be merged.

CookStar:
I added a fix to 9c7e4a22d5d465f705379457d5ce704ca0aa3eb6.
Fixed Player.is_bot to be cached correctly. #294

Apparently it seems to crash when unloading Source.Python if these codes are loaded.
```python
# Source.Python
#   Engines
from engines.server import engine_server
#   Memory
from memory import get_virtual_function
from memory.hooks import PreHook

@PreHook(get_virtual_function(engine_server, "ServerCommand"))
def pre_server_command(args):
    pass
```
Output: 
```
quit
Host_ShutdownServer
CBaseServer::Shutdown
CSteam3Server::Shutdown
NET_SteamGameServer_Shutdown
SteamGameServer_Shutdown
SV_ShutdownGameDLL
[Source.Python] Unloading...
free(): invalid pointer
```

jordanbriere:
> 
> 
> I added a fix to [9c7e4a2](https://github.com/Source-Python-Dev-Team/Source.Python/commit/9c7e4a22d5d465f705379457d5ce704ca0aa3eb6).
> Fixed Player.is_bot to be cached correctly. #294
> 

Nice catch, thanks!

> Apparently it seems to crash when unloading Source.Python if these codes are loaded.
> 
> ```python
> # Source.Python
> #   Engines
> from engines.server import engine_server
> #   Memory
> from memory import get_virtual_function
> from memory.hooks import PreHook
> 
> @PreHook(get_virtual_function(engine_server, "ServerCommand"))
> def pre_server_command(args):
>     pass
> ```
> 
> Output:
> 
> ```
> quit
> Host_ShutdownServer
> CBaseServer::Shutdown
> CSteam3Server::Shutdown
> NET_SteamGameServer_Shutdown
> SteamGameServer_Shutdown
> SV_ShutdownGameDLL
> [Source.Python] Unloading...
> free(): invalid pointer
> ```

I haven't look into it yet, but my first guess would be [here](https://github.com/Source-Python-Dev-Team/Source.Python/blob/cached_property/src/core/modules/memory/memory_function.cpp#L195-L203). By the time the functions are deleted, DynamicHooks already uninstalled all the hooks and handled the conventions so that condition passes and we then attempt to free an already freed pointer.

jordanbriere:
I've finally had a chance to look into that crash further and it was indeed caused by what I mentioned above. I've added a patch for that into https://github.com/Source-Python-Dev-Team/Source.Python/pull/292/commits/debc63bf4a641476014b6105b6a333e608645410, but I still believe DynamicHooks should ideally not free something it didn't allocate. Perhaps using a `shared_ptr` could be smarter to ensure the conventions are freed when everyone is done with them. Also, that means we are now leaking as nobody will be handling the convention anymore if `CFunction::DeleteHook` is called.

Frag1337:
Is there any reason or info why it is not added to the main branch? Are there any conflicts?

jordanbriere:
> 
> 
> Is there any reason or info why it is not added to the main branch? Are there any conflicts?

Although everything seems to be working fine from my testings, that branch contains some drastic changes to various internal mechanisms that could and most certainly does hide underlying issues and there is so much I can locally test, and even more I can overlook. 😄

If you wish to help it progress, then the easiest way would be to play around with it and report any issues you may find. Based on your recent issues, I've pre-compiled a build for CS:S/Windows that you can download there: [cached_property-css-windows.zip](https://drive.google.com/file/d/1WfLhB7KzL9UQscnoRLmy7PG6xEnzwrA3/view)

Here is a summary of the changes it contains and what to look for issues (along with everything that gravitate around those and could potentially be affected):

```
Added cached properties implementation.
Entity's dynamic functions, inputs, outputs, and various properties are now cached.
Improved performance of Entity.<get/set>_property_<type> methods.
Fixed memory leaks related to Function's convention.
Added segfault handler to Binary.find_pointer.
Fixed some exceptions being silenced under certain conditions within the load/unload process.
Added BaseEntity.entity_flags.
Fixed an issue when entities instances were cached again after we invalidated the cache.
Fixed EntityDictionary potentially caching soon-to-be removed entities.
Some various improvements to some redundant logics.
Entity.__hash__ now hashes its inthandle rather than its index.
Fixed the invalidation of the internal entity cache before all entity deletion listeners were called.
Fixed a KeyError being silenced in EntityDictionary.on_automatically_removed overrides.
Slightly improved the retrieval speed of the cached entity instances.
Improved performance of memory tools (approximately 2.5 times faster).
```
If you find any issues, please post directly here instead of opening tracked issues. Thanks!

addons/source-python/packages/source-python/entities/_base.py
```diff
@@ -772,7 +910,10 @@ def set_property_vector(self, name, value):
         :param Vector value:
             The value to set.
         """
-        self._set_property(name, 'Vector', value)
+        try:
+            self.set_network_property_vector(name, value)
+        except ValueError:
+            self.set_datamap_property_vector(name, value)
```
Ayuto, line 916: Shouldn't we do the same with ``Edict`` and then remove ``_get_property`` and ``_set_property``. That seems to be the last one that uses the slow variant by iterating over the server classes and datamaps.

addons/source-python/packages/source-python/players/_base.py
```diff
@@ -221,20 +217,36 @@ def is_fake_client(self):
         """
         return self.playerinfo.is_fake_client()
 
-    def is_hltv(self):
+    @cached_property
+    def _is_hltv(self):
```
Ayuto, line 221: Why is this wrapper needed?

addons/source-python/packages/source-python/players/_base.py
```diff
@@ -262,7 +274,7 @@ def set_team(self, value):
 
     team = property(get_team, set_team)
 
-    @property
+    @cached_property
```
Ayuto, line 277: Theoretically, the client's language can be changed. So, I'm not sure if we should cache it.

src/core/modules/core/core_cache.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+void CCachedProperty::__set_name__(object owner, str name)
+{
+	m_name = name;
+	m_owner = object(handle<>(PyWeakref_NewRef(owner.ptr(), NULL)));
+}
+
+object CCachedProperty::__get__(object self, object instance, object owner=object())
+{
+	CCachedProperty &pSelf = extract<CCachedProperty &>(self);
```
Ayuto, line 211: Should we move this line a little bit down (behind the check), since ``pSelf`` isn't used if ``instance`` is ``None``? Or is it done here to let ``extract`` do some validations?

addons/source-python/packages/source-python/entities/_base.py
jordanbriere: Ideally, yes. But `BaseEntity.get_<network/datamap>_property_edict` are not implemented and as far as dumping goes, there isn't any properties that are edicts (at least, on OB games) to implement them and base tests upon. So they were left as is, although I think they could probably be removed. Not even sure they are `edict_t *` as the current implementation assume, and they could well be anything that require to be proxied as far as I know.

addons/source-python/packages/source-python/players/_base.py
jordanbriere: This would break compatibility, because they are currently not properties but functions. So existing codes are doing `pl.is_<bot/hltv>()` and not having them wrapped would means they would now get a `bool isn't callable` exception.

addons/source-python/packages/source-python/players/_base.py
jordanbriere: True, but since it is already cached [here](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/players/_language/cache.py) I think it is overall beneficial to only query that external dictionary once and bind the result.

src/core/modules/core/core_cache.cpp
jordanbriere: No specific reason why it is done there, and I agree, it should be moved and casted only when needed.

addons/source-python/packages/source-python/players/_base.py
Ayuto: Ah, I didn't thought about that. Thanks for the clarification.

jordanbriere:
I know a few people tested that branch and since nothing new was reported I assume it works fine for everyone. Any other points you would like to be addressed @Ayuto? If not, let's get this merged and address potential issues in due time.

Ayuto:
No, I don't have anything to add :)

PR #295, opened 2019-12-28
Entity.repeat

invisiblesoldiers:
Added analogue for Entity.delay but for Repeat.
```python

def print_alive(player):
    print('Alive')

player = Player(1)

# Will be stopped after removing the entity.
repeat = player.register_repeat(print_alive, (player, ))

# You still can manually stop the repeat.
if player.dead:
    repeat.stop() 
```
Garbage collecting for args of Repeat class.
```python
def __del__(self)
```
never will be called in a Entity subclass if you will pass reference to the instance of Entity object.  
The object remains in the memory forever.

For example:
```python
from random import random

class CustomPlayer(Player):
    def __init__(self, index):
        super().__init__(index)
   
        # passed function contains reference to player
        repeat = self.register_repeat(self._print_random_handler)

    # Never will be called.
    def __del__(self):
        print('__del__')

    def _print_random_handler(self):
        print(random())
          
PLAYERS = PlayerDictionary(CustomPlayer)
PLAYERS[1]
```
I barely deal with it using _weakref_ library. But there are still gaps. For example you still can pass the reference like additional argument and again the object never will be deleted. I'm waiting for comments about it.
```python
repeat = self.register_repeat(self._print_random_handler, (self, ))
```

### DIFF
--- a/addons/source-python/packages/source-python/entities/_base.py
+++ b/addons/source-python/packages/source-python/entities/_base.py
@@ -50,6 +50,8 @@ from filters.weapons import WeaponClassIter
 from listeners import OnEntityDeleted
 from listeners import on_entity_deleted_listener_manager
 from listeners.tick import Delay
+from listeners.tick import Repeat
+from listeners.tick import RepeatStatus
 #   Mathlib
 from mathlib import NULL_VECTOR
 #   Memory
@@ -69,6 +71,9 @@ _projectile_weapons = [weapon.name for weapon in WeaponClassIter('grenade')]
 
 # Get a dictionary to store the delays
 _entity_delays = defaultdict(set)
+
+# Get a dictionary to store the repeats
+_entity_repeats = defaultdict(set)
 
 # Get a set to store the registered entity classes
 _entity_classes = WeakSet()
@@ -817,6 +822,29 @@ class Entity(BaseEntity, metaclass=_EntityCaching):
         # Return the delay instance...
         return delay
 
+    def register_repeat(self, callback, args=(), kwargs=None):
+        """Register a repeat which will be deleted and canceled after removing the entity.
+
+        :param callback:
+            A callable object that should be called at the end of each loop.
+        :param tuple args:
+            Arguments that should be passed to the callback.
+        :param dict kwargs:
+            Keyword arguments that should be passed to the callback.
+        :raise ValueError:
+            Raised if the given callback is not callable.
+        """
+
+        # Get the repeat instance...
+        repeat = Repeat(_callback, args, kwargs)
+
+        # Add the repeat to the dictionary...
+        _entity_repeats[self.index].add(repeat)
+
+        # Return the repeat instance...
+        return repeat
+
+
     def get_input(self, name):
         """Return the input function matching the given name.
 
@@ -1103,19 +1131,30 @@ def _on_entity_deleted(base_entity):
     for cls in _entity_classes:
         cls.cache.pop(index, None)
 
-    # Was no delay registered for this entity?
-    if index not in _entity_delays:
-        return
+    # Was delay registered for this entity?
+    if index in _entity_delays:
+        # Loop through all delays...
+        for delay in _entity_delays[index]:
 
-    # Loop through all delays...
-    for delay in _entity_delays[index]:
+            # Make sure the delay is still running...
+            if not delay.running:
+                continue
 
-        # Make sure the delay is still running...
-        if not delay.running:
-            continue
+            # Cancel the delay...
+            delay.cancel()
 
-        # Cancel the delay...
-        delay.cancel()
+        # Remove the entity from the dictionary...
+        del _entity_delays[index]
 
-    # Remove the entity from the dictionary...
-    del _entity_delays[index]
+    # Was repeat registered for this entity?
+    if index in _entity_repeats:
+        # Loop through all repeats...
+        for repeat in entity_repeats[index]:
+
+            # Stop the repeat if running
+            if repeat.status is RepeatStatus.RUNNING:
+                repeat.stop()
+
+        # Remove the entity from the dictionary...
+        del _entity_repeats[index]
+            
--- a/addons/source-python/packages/source-python/entities/_base.py
+++ b/addons/source-python/packages/source-python/entities/_base.py
@@ -779,7 +779,7 @@ class Entity(BaseEntity, metaclass=_EntityCaching):
     def delay(
             self, delay, callback, args=(), kwargs=None,
             cancel_on_level_end=False):
-        """Execute a callback after the given delay.
+        """Create the delay which will be stopped after removing the entity.
 
         :param float delay:
             The delay in seconds.
@@ -822,8 +822,8 @@ class Entity(BaseEntity, metaclass=_EntityCaching):
         # Return the delay instance...
         return delay
 
-    def repeat(self, callback, args=(), kwargs=None):
-        """Create a repeat which will be stopped after removing the entity.
+    def repeat(self, callback, args=(), kwargs=None, cancel_on_level_end=False):
+        """Create the repeat which will be stopped after removing the entity.
 
         :param callback:
             A callable object that should be called at the end of each loop.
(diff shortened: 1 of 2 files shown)


invisiblesoldiers:
I noticed that the weakrefs block building a menu after 1.0s

jordanbriere:
Nice! Here are few points:

- This [line](https://github.com/Source-Python-Dev-Team/Source.Python/pull/295/files#diff-b816a3fa386e9526bc4f234138e782abR839) will raise because `_callback` is not defined. Should be `callback`, per your prototype.
- This [line](https://github.com/Source-Python-Dev-Team/Source.Python/pull/295/files#diff-b816a3fa386e9526bc4f234138e782abR1152) will raise because `entity_repeats` is not defined. Should be `_entity_delays`.
- Using weak references [here](https://github.com/Source-Python-Dev-Team/Source.Python/pull/295/files#diff-11b8334e6500401ad6339b84b1ec1ad3R240-R242) is a bad idea. If a `lambda` is used as a callback, or if the keywords dictionary is passed in-line (e.g. `Repeat(callback, kwargs={'foo': 'bar'}`), they won't live long enough which will raise when it is time to process them. You also didn't explicitly import `WeakValueDictionary`, meaning it will raise as well.
- `Entity.register_repeat` should really be named `Entity.repeat`, to be consistent with `Entity.delay`.

invisiblesoldiers:
> * This [line](https://github.com/Source-Python-Dev-Team/Source.Python/pull/295/files#diff-b816a3fa386e9526bc4f234138e782abR839) will raise because `_callback` is not defined. Should be `callback`, per your prototype.
> 
> * This [line](https://github.com/Source-Python-Dev-Team/Source.Python/pull/295/files#diff-b816a3fa386e9526bc4f234138e782abR1152) will raise because `entity_repeats` is not defined. Should be `_entity_repeats`.
Oh, apparently I downloaded the old version, I've already fixed it

invisiblesoldiers:
> * Using weak references [here](https://github.com/Source-Python-Dev-Team/Source.Python/pull/295/files#diff-11b8334e6500401ad6339b84b1ec1ad3R240-R242) is a bad idea. If a `lambda` is used as a callback, or if the keywords dictionary is passed in-line (e.g. `Repeat(callback, kwargs={'foo': 'bar'}`), they won't live long enough which will raise when it is time to process them. You also didn't explicitly import `WeakValueDictionary`, meaning it will raise as well.
>
Then what we can do with garbage collecting without weakrefs? Or so it should be.

invisiblesoldiers:
For example, i have the real code
```python
class TimerPlayer(Player, TrackAttributes):
    def __init__(self, index):
        Player.__init__(self, index, caching=False)
        TrackAttributes.__init__(self)

    def on_automatically_removed(self):
        TrackAttributes.on_automatically_removed(self)

    def __del__(self):
        print('__del__')
```
```python
class TrackAttributes(object):
    def __init__(self):
        self.creating_track = None
        self.creating_zone = None
        self.selected_track = None
        self.selected_zone = None
        self.selected_point = ZonePoint.LOWER

        self.gridvalue = 64
        self.pointmethod = PointMethod.POSITION

        self.zonecategory = ZoneCategory.BASIS
        self.trackinfo_lastmenu = None
        self.zoneinfo_lastmenu = None

        self.gridsnapping_view = Repeat(self._gridsnapping_view_handler)

    def on_automatically_removed(self):
        if self.gridsnapping_view.status is RepeatStatus.RUNNING:
            self.gridsnapping_view.stop()
            
        # Clear the reference to player
        self.gridsnapping_view.callback = None
```
And _del_ will be called because the line here
self.gridsnapping_view.callback = None
I cleared a reference to the player.

jordanbriere:
> And _del_ will be called because the line here
> self.gridsnapping_view.callback = None
> I cleared a reference to the player.

This is not an issue. The traverse mechanism will catch up later and delete that object. An object is deleted when its reference count reaches zero, or when it is traversed by the garbage collector in case of circular references. The collection is done periodically, but you can force it with:

```python
from gc import collect

collect()
```

Basically, when an object reaches a certain reference count, Python marks it as potentially collectible, will traverse it during its next run and will delete it if the count was being held hostage due to a circular reference. You do not have to worry about any of this yourself. I assume the paranoia has been created by me into https://github.com/Source-Python-Dev-Team/Source.Python/pull/292, but the difference there is the objects are proxies of pointers owned by Boost, preventing Python from freeing them even when traversed by lack of ownership.

In conclusion; unless under very specific circumstances, you should not have to worry about circular references and there is nothing wrong with the current implementation of Repeat storing strong references to the callbacks and arguments.

invisiblesoldiers:
>  jordanbriere 
>
Well, thank you for reassuring me. So I can only offer _Entity.repeat_

invisiblesoldiers:
Also i removed _cancel_on_level_end_ argument to _Delay_. Because as I know _@OnEntityDeleted_ will be called on level end anyway.

addons/source-python/packages/source-python/entities/_base.py
```diff
@@ -817,6 +822,29 @@ def _callback(*args, **kwargs):
...
+        # Get the repeat instance...
+        repeat = Repeat(callback, args, kwargs)
+
+        # Add the repeat to the dictionary...
+        _entity_repeats[self.index].add(repeat)
+
+        # Return the repeat instance...
+        return repeat
+
+
```
satoon101: There should only be 1 blank line before a method, not 2.

jordanbriere:
> 
> 
> Also i removed _cancel_on_level_end_ argument to _Delay_. Because as I know _@OnEntityDeleted_ will be called on level end anyway.

Technically, yes. But that should not be our call to assume what the caller wants to forward to the constructor. I mean, these are effectively wrappers of the `Delay` and `Repeat` classes, meaning their prototype should mirror in my opinion. Moreover, and I know this is your favourite statement... backward compatibility! :grimacing:

invisiblesoldiers:
> > Also i removed _cancel_on_level_end_ argument to _Delay_. Because as I know _@OnEntityDeleted_ will be called on level end anyway.
> 
> Technically, yes. But that should not be our call to assume what the caller wants to forward to the constructor. I mean, these are effectively wrappers of the `Delay` and `Repeat` classes, meaning their prototype should mirror in my opinion. Moreover, and I know this is your favourite statement... backward compatibility! grimacing
>
Why? Do you reference to the conflict with _PlayerDictionary_ and _Player.cache_. This is different.  The repeat will be linked with it's entity. And I just made it easier for someone to create a _Repeat_ for the case when it must be stopped after entity deleted. Perhaps it is necessary to explain in the comments on the wiki that level end lead to removing the entitiy.  
If anyone wants a more complicated, there is vanila _Repeat_.  
I'll wait for your answer and maybe you'll agree and I'll put it away the same way in Entity.delay.  
If not, I'll add back.

jordanbriere:
The point I'm trying to make is, `Entity.delay` and `Entity.repeat` are primarily just wrapper methods used to instantiate `Delay` and `Repeat` objects respectively. Being wrappers, they simply forward the arguments and keywords they are being fed to their respective class constructor. For example, the `Delay` prototype is the following:

```python
Delay(delay, callback, args=(), kwargs=None, cancel_on_level_end=False)
```

So a wrapper of this class, such as `Entity.delay`, or a wrapper of `Repeat` such as `Entity.repeat` should ideally mirror that prototype and accept every arguments and keywords that can possibly be forwarded to the class being wrapped because at the end of the day, this is a wrapper used to instantiate an object of an existing class, and this is not our call to decide whether or not we should strip arguments and keywords from that instantiation and that should be up to the users to decide.

Also, while this is true that entities are deleted on map changes, calling `entity.delay(10, whatever, cancel_on_level_end=True)` will have different behaviour than `cancel_on_level_end=False` because the entities are deleted slightly after the map has officially ended.

I was referring to backward compatibility, because by removing that keyword, this changes the prototype of an existing method, and there is always possibilities of someone somewhere using it causing their codes to now need updating. Here are the cons of such change:

- Breaks existing code that might use that keyword for whatever reasons.
- Removes available flexibility of the objects users are instantiating.

Please, enlighten me of the benefits of such change, as I can't think of any?

invisiblesoldiers:
I will move my text upper to #292. And i have a solution to kill two birds with one stone.

PR #297, opened 2020-01-04
Player.from_userid caching

satoon101:
* Added caching boolean argument for Player.from_userid.

### DIFF
--- a/addons/source-python/packages/source-python/entities/_base.py
+++ b/addons/source-python/packages/source-python/entities/_base.py
@@ -89,11 +89,13 @@ class _EntityCaching(BoostPythonClass):
 
         # Set whether or not this class is caching its instances by default
         try:
-            cls._caching = signature(
-                cls.__init__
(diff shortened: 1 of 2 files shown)


satoon101:
This came about because I was testing GunGame this morning and encountered an issue with gg_multi_level.  I narrowed down the issue to the fact that the cache was causing me to get the same instance of _MultiLevelPlayer every time.  There were other ways to fix it and use the cache, but I at least thought it would be a good idea to allow for caching=False through from_userid.

jordanbriere:
Yes, makes sense for that method to forward that keyword. We should also add it to `Entity.from_inthandle` as well. Since it breaks compatibility, I believe we should default it to `False` here [here](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/entities/_base.py#L96) instead of `True` so that existing subclasses that were not implemented with caching in mind continues to works as they used to.

PR #318, opened 2020-04-22
Added CCrossbowBolt::BoltTouch signature for HL2DM.

vinci6k:


jordanbriere:
Thanks! However I don't think it would be wise to add an extra signature to maintain for that function. I proposed an alternative in the [thread that encouraged this PR](https://forums.sourcepython.com/viewtopic.php?f=37&t=2231&p=13855#p13855), and if that function was specifically needed to be hooked, I think it would be overall more robust and maintainable to dynamically grab it through `m_pfnTouch` that is available in the `crossbow_bolt`'s datamap which would eliminate the need to keep data up-to-date for it (even if I doubt that game will be updated anytime soon 😄).

Ayuto:
I agree. That function is a very special case and I would like to keep the number of signatures and offsets as less as possible.

vinci6k:
Understandable. Guess I jumped the gun on this one.
> I think it would be overall more robust and maintainable to dynamically grab it through `m_pfnTouch` that is available in the `crossbow_bolt`'s datamap which would eliminate the need to keep data up-to-date for it (even if I doubt that game will be updated anytime soon 😄).

If it's not too much of a hassle, could you give me an example on how to do that?

jordanbriere:
> If it's not too much of a hassle, could you give me an example on how to do that?

Sure, here is a basic example:

```python
# Create a bolt
bolt = Entity.create('crossbow_bolt')

# The touch function is bound when the bolt is spawned
bolt.spawn()

# Grab the touch function
bolt_touch = bolt.get_datamap_property_pointer('m_pfnTouch').make_function(
    Convention.THISCALL,
    (DataType.POINTER, DataType.POINTER),
    DataType.VOID
)

# Do whatever with it
print('CCrossbowBolt::Touch\'s address is:', bolt_touch.address)
```

vinci6k:
Awesome, thanks!

PR #319, opened 2020-04-22
Added score to CS:GO data.

CookStar:


vinci6k:
Oooh, awesome! I needed this offset for a plugin [ages ago](https://forums.sourcepython.com/viewtopic.php?f=20&t=1642).

PR #322, opened 2020-04-27
Fixed logger outputting a duplicate message after reloading plugins.

CookStar:
This fixes the logger's handler remaining after a plugin reload, resulting in a duplicate message.

Test Code.
```python
# ../test/test.py

# Source.Python
#   Commands
from commands.typed import TypedServerCommand
#   Cvars
from cvars import ConVar
#   Loggers
from loggers import LogManager

level = ConVar("test_logging_level", "0", "Test logging level.")
areas = ConVar("test_logging_areas", "9", "Test logging areas.")

test_logger = LogManager(
    "test",
    level,
    areas,
    "test",
    "%(asctime)s - %(name)s - %(message)s",
    "%Y-%m-%d %H:%M:%S"
)

@TypedServerCommand("test")
def on_test(command_info):
    test_logger.log_message("Test Log")
```
Input:
```
sp plugin reload test
test
```
Output(Before):
```
2020-04-27 11:18:35 - test - Test Log
2020-04-27 11:18:35 - test - Test Log
```
Output(After):
```
2020-04-27 11:20:53 - test - Test Log
```


addons/source-python/packages/source-python/loggers.py
```diff
@@ -388,6 +390,10 @@ def areas(self):
         """
         return self._areas.get_int()
 
+    def _unload_instance(self):
+        """Remove the handler from the logger."""
+        self.logger.removeHandler(self._handler)
```
jordanbriere, line 395: Wouldn't it be better to delete the logger itself?
```python
self.logger.manager.loggerDict.pop(self.logger.name, None)
```
This would ensure handlers registered through `test_logger.logger.addHandler(...)` directly also get removed.

CookStar: That's certainly the appropriate way to do it!

PR #324, opened 2020-04-28
Fixed AttributeError when displaying plugin list.

CookStar:


Ayuto:
Which issue does this solve? Could you please add a few steps to reproduce the issue?

jordanbriere:
> 
> 
> Which issue does this solve? Could you please add a few steps to reproduce the issue?

It cause an error if it is set to `False` and wasn't overwritten by a `ConVar` on plugin load. Basically, it will never explicitely be `None`, either `False` or a `ConVar` if it was `True` when the plugin was loaded (which should always be the case for iterated plugins by `sp plugin list`). If you have the following in your `info.ini`:
```ini
public_convar = 0
```
Then the following error will raise when using `sp plugin list`:
```python
AttributeError: 'bool' object has no attribute 'name'
```
However, a proper fix should rather be:
```python
            if isinstance(info.public_convar, ConVar):
```
Because we implemented [`ConVar.__bool__`](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/cvars/cvars_wrap.cpp#L316-L320) which will evaluates to `False` if the value cannot be interpreted as `1`/`0` resulting into the variables no longer be listed in the output. For example, this would prints `False`, although is a valid versioning variable:

```python
print(bool(ConVar('foo', '0.1')))
```

CookStar:
> Which issue does this solve? Could you please add a few steps to reproduce the issue?

I apologize for the lack of explanation.


> It cause an error if it is set to `False` and wasn't overwritten by a `ConVar` on plugin load. Basically, it will never explicitely be `None`, either `False` or a `ConVar` if it was `True` when the plugin was loaded (which should always be the case for iterated plugins by `sp plugin list`). If you have the following in your `info.ini`:
> 
> ```ini
> public_convar = 0
> ```
> 
> Then the following error will raise when using `sp plugin list`:
> 
> ```python
> AttributeError: 'bool' object has no attribute 'name'
> ```

As jordanbriere wrote above, that's the problem.


> However, a proper fix should rather be:
> 
> ```python
>             if isinstance(info.public_convar, ConVar):
> ```
> 
> Because we implemented [`ConVar.__bool__`](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/cvars/cvars_wrap.cpp#L316-L320) which will evaluates to `False` if the value cannot be interpreted as `1`/`0` resulting into the variables no longer be listed in the output. For example, this would prints `False`, although is a valid versioning variable:
> 
> ```python
> print(bool(ConVar('foo', '0.1')))
> ```

Oh, I didn't know about ConVar's behavior.
Changed to check PublicConVar's classinfo.

Issue #236, opened 2018-01-14
take_damage() with skip_hooks=True doesn't work as expected


Ayuto:
``skip_hooks=True`` doesn't work as expected, because ``on_take_damage`` is a virtual function. Example:
1. Hook ``on_take_damage`` on a human player.
2. Call ``take_damage`` with ``skip_hooks=True`` on a bot.

The hook will still get called, because bots have their own implementation (CCSBot) of ``on_take_damage`` which later calls the overwritten implementation (CCSPlayer).

One workaround would be to create an ``OnTakeDamage`` listener and let Source.Python call the listener from an internal hook. Here's some pseudo-code:
```python
# Internal SP code
disabled = False
    
def internal_hook(args):
    if disabled:
        return

    OnTakeDamage.manager.notify(...)
    
def take_damage(self, ..., skip_hooks=True): # Let it default to True
    global disabled
    disabled = skip_hooks
    
    # Use try/finally just to be sure
    try:
        really_do_damage()
    finally:
        disabled = False
```
```python
# Plugin code
@OnTakeDamage
def on_take_damage(entity, info):
    # You can call it directly
    take_damage()

    # Or delayed
    delay(10, take_damage)

# Or from somewhere else
take_damage()
```
The listener would also make hooking ``on_take_damage`` much easier for plugin authors.

PR #326, opened 2020-04-30
Added the ability to disable hooks

Ayuto:
Generic fix for issue #236

Test code:
```python
from memory.hooks import hooks_disabled
from entities.hooks import EntityPreHook, EntityCondition
from events import Event

@EntityPreHook(EntityCondition.is_human_player, 'on_take_damage')
def pre_take_damage(args):
    print('take damage', args)

@Event('player_say')
def on_player_say(event):
    player = Player.from_userid(event['userid'])
    target = player.view_player
    if target is None:
        print('not a player or no target')
        return

    with hooks_disabled(event['text'] == 'yes'):
        target.take_damage(5)
```
When looking at a player and typing ``yes``, the hook shouldn't get called. In all other cases the hook gets called.

### DIFF
--- a/src/core/modules/memory/memory_hooks.cpp
+++ b/src/core/modules/memory/memory_hooks.cpp
@@ -44,6 +44,8 @@ using namespace boost::python;
 // g_mapCallbacks[<CHook *>][<HookType_t>] -> [<object>, <object>, ...]
 std::map<CHook *, std::map<HookType_t, std::list<object> > > g_mapCallbacks;
 
+bool g_HooksDisabled;
+
 
 // ============================================================================
 // >> HELPER FUNCTIONS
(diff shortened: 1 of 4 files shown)


src/core/modules/memory/memory_hooks.cpp
```diff
@@ -83,7 +85,7 @@ bool SP_HookHandler(HookType_t eHookType, CHook* pHook)
 	std::list<object> callbacks = g_mapCallbacks[pHook][eHookType];
 
 	// No need to do all this stuff, if there is no callback registered
-	if (callbacks.empty())
+	if (g_HooksDisabled || callbacks.empty())
```
jordanbriere, line 88: Not really dramatic but, since there is no need to query the map and initialize `callbacks` if the hooks are globally disabled, I think it would be better to move that check at the very top.

Ayuto: Good point! I will do that later.

PR #337, opened 2020-06-16
Modified player.get_godmode() for CSGO.

vinci6k:
Godmode granted through spawn protection or other sources within official gamemodes (deathmatch, arms race) isn't covered by the **m_takedamage** property. With this change, checking if a player has godmode in CSGO should always work.


Ayuto:
Thank you!

PR #338, opened 2020-06-20
Update pymysql

Frag1337:


Frag1337:
When connecting to MySQL 8, pymysql may raise an exception, such as KeyError: 255 due to new charsets added.

So I've upgraded the package to the most recent version, which fixes the issue.

Ayuto:
Thank you!

PR #340, opened 2020-06-29
Add TF2 specific player subclass. Add class names and id's to team data

ns-phennessy:
Per this thread:
https://forums.sourcepython.com/viewtopic.php?f=10&t=2320&p=14295#p14295

It feels a bit clunky to have to load the config file directly into my class; I figured this data would be available some other way?

### DIFF
--- a/addons/source-python/packages/source-python/players/engines/orangebox/tf.py
+++ b/addons/source-python/packages/source-python/players/engines/orangebox/tf.py
@@ -1,0 +1,50 @@
+# ../players/engines/orangebox/cstrike.py
+
+"""Provides Team Fortress 2 specific Player based functionality."""
+
+# =============================================================================
+# >> IMPORTS
+# =============================================================================
+# Source.Python Imports
+from . import Player as _Player
+from configobj import ConfigObj
+from core import GAME_NAME
+from paths import SP_DATA_PATH
+
+# =============================================================================
+# >> CLASSES
+# =============================================================================
+class Player(_Player):
+    caching = True
+    
+    # Team
+    _team_data = ConfigObj(SP_DATA_PATH / 'teams' / GAME_NAME + '.ini', unrepr=True)
+    teams_by_name = _team_data.get('names')
+    teams_by_number = {number: alias for alias, number in teams_by_name.items()}
+
+    @property
+    def team_name(self):
+        return self.teams_by_number[self.team]
+    
+    @team_name.setter
+    def team_name(self, team_name):
+        if team_id not in self.team_by_name.keys():
+            raise ValueError(f"Invalid team_name: {team_name}")
+        
+        self.team = self.teams_by_name[team_name]
+
+    # Mercanary class
+    class_by_name = _team_data.get('classes')
+    class_by_number = {number: alias for alias, number in class_by_name.items()}
+    
+    @property
+    def player_class_name(self):
+        return self.class_by_number[self.player_class]
+
+    @player_class_name.setter
+    def player_class_name(self, class_name):
+        if team_id not in self.class_by_name.keys():
+            raise ValueError(f"Invalid class_name: {class_name}")
+        
+        # Does this need to be desired_player_class?
+        self.player_class = self.class_by_name[team_name]
--- a/addons/source-python/packages/source-python/players/engines/orangebox/tf.py
+++ b/addons/source-python/packages/source-python/players/engines/orangebox/tf.py
@@ -30,7 +30,7 @@ class Player(_Player):
     
     @team_name.setter
     def team_name(self, team_name):
-        if team_id not in self._teams_names:
+        if team_name not in self._teams_names:
             raise ValueError(f"Invalid team_name: {team_name}")
         
         self.team = self._teams_names[team_name]
@@ -42,8 +42,8 @@ class Player(_Player):
 
     @player_class_name.setter
     def player_class_name(self, class_name):
-        if team_id not in self._class_names:
+        if class_name not in self._class_names:
             raise ValueError(f"Invalid class_name: {class_name}")
         
-        # Does this need to be desired_player_class or should there be a setter for desired class?
-        self.player_class = self._class_names[team_name]
+        # Does this need to be desired_player_class or should there b a setter for desired class?
+        self.player_class = self._class_names[class_name]
--- a/addons/source-python/packages/source-python/players/engines/orangebox/tf.py
+++ b/addons/source-python/packages/source-python/players/engines/orangebox/tf.py
@@ -19,32 +19,31 @@ class Player(_Player):
     
     # Team
     _team_data = ConfigObj(SP_DATA_PATH / 'teams' / GAME_NAME + '.ini', unrepr=True)
-    teams_by_name = _team_data.get('names')
-    teams_by_number = {number: alias for alias, number in teams_by_name.items()}
+
+    _teams_names = _team_data.get('names')
+    _class_names = _team_data.get('classes')
 
     @property
     def team_name(self):
-        return self.teams_by_number[self.team]
+        teams_by_number = dict(zip(self._teams_names.values(), self._teams_names.keys()))
+        return teams_by_number[self.team]
     
     @team_name.setter
     def team_name(self, team_name):
-        if team_id not in self.team_by_name.keys():
+        if team_id not in self._teams_names:
             raise ValueError(f"Invalid team_name: {team_name}")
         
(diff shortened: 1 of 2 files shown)


ns-phennessy:
Anything I can to do help test this so it can be merged? 

I have run this on one of my own servers just fine, and works as expected

addons/source-python/packages/source-python/players/engines/orangebox/tf.py
```diff
@@ -0,0 +1,49 @@
...
+from paths import SP_DATA_PATH
+
+# =============================================================================
+# >> CLASSES
+# =============================================================================
+class Player(_Player):
+    caching = True
+    
+    # Team
+    _team_data = ConfigObj(SP_DATA_PATH / 'teams' / GAME_NAME + '.ini', unrepr=True)
```
jordanbriere, line 21: There is no need to parse that file again here, as it is already parsed into [../players/teams.py](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/players/teams.py#L31-L33).

addons/source-python/packages/source-python/players/engines/orangebox/tf.py
```diff
@@ -0,0 +1,49 @@
...
+
+    _teams_names = _team_data.get('names')
```
jordanbriere, line 23: This mapping is already available as [players.teams.teams_by_name](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/players/teams.py#L35-L36) and should really just be imported from there rather than queried again from the INI file.

addons/source-python/packages/source-python/players/engines/orangebox/tf.py
```diff
@@ -0,0 +1,49 @@
...
+    _class_names = _team_data.get('classes')
+
+    @property
+    def team_name(self):
+        teams_by_number = dict(zip(self._teams_names.values(), self._teams_names.keys()))
+        return teams_by_number[self.team]
+    
+    @team_name.setter
```
jordanbriere, line 31: Rather than having multiples `team_name` methods decorated as getter and setter I think it would be better to follow the overall design of the other properties across the whole project where they are distinctly available as methods as well. For example, `get_team_name` and `set_team_name` bound to the `team_name` property.

addons/source-python/packages/source-python/players/engines/orangebox/tf.py
```diff
@@ -0,0 +1,49 @@
...
+    def team_name(self, team_name):
+        if team_name not in self._teams_names:
+            raise ValueError(f"Invalid team_name: {team_name}")
+        
+        self.team = self._teams_names[team_name]
+    
+    @property
+    def player_class_name(self):
+        class_by_number = dict(zip(self._class_names.values(), self._class_names.keys()))
```
jordanbriere, line 40: This is redundant to rebuild that mapping every calls, and it should really just be globalized into the `players.teams` module like the teams mapping are.

addons/source-python/packages/source-python/players/engines/orangebox/tf.py
```diff
@@ -0,0 +1,49 @@
...
+        return class_by_number[self.player_class]
+
+    @player_class_name.setter
+    def player_class_name(self, class_name):
+        if class_name not in self._class_names:
```
jordanbriere, line 45: The dictionary here is queried twice (once through `__contains__`, then another through `__getitem__`). A good practice is to query the dictionary only once, for example using `dict.get`:

```py
v = d.get(k, None)

# If k wasn't in d, let's raise
if v is None:
    raise

...
```
This would ensure the dictionary is looked up only once rather than twice, etc.

ns-phennessy:
Awesomse thanks! Ill make some changes based on your comments. 

> That said, rather than being implemented solely for TF2, it would be good to implement that for other games that support classes as well. An implementation that simply raise a `NotImplementedError` for games that are not supported or that don't have classes data in their file would be good enough.

How would that work? Would you rather see this just on the Player base class instead? I was told in [this thread](https://forums.sourcepython.com/viewtopic.php?f=10&t=2320&p=14295#p14295) to make a tf2 specific player subclass. Additionally I would like to add other TF2 specific things to this over time like weapon data and such. I don't have a strong opinion on it, just unsure how you want it to look. 

Thanks!

jordanbriere:
> How would that work? Would you rather see this just on the Player base class instead? I was told in [this thread](https://forums.sourcepython.com/viewtopic.php?f=10&t=2320&p=14295#p14295) to make a tf2 specific player subclass. Additionally I would like to add other TF2 specific things to this over time like weapon data and such. I don't have a strong opinion on it, just unsure how you want it to look.

Having a TF2's specific Player class is good, for what is actually specific for that game. When it comes to classes, many (if not all) supported games works the same way as TF2 does so it makes sense to have that functionality available in the base Player class. For example, on CS:S the classes are:

```
Phoenix Connexion = 1
Elite Crew = 2
Artic Avengers = 3
Guerilla Warfare = 4
Seal Team 6 = 5
GSG-9 = 6
SAS = 7
GIGN = 8
```

Rather than raising when no data is available, perhaps it would be easier to just return default values such as class index `-1` and class name `Unknown`. That way there would not have any data validation to be done, as random `KeyError`, etc. wouldn't be explicit enough for games that aren't supported. I made a quick test just now with the following on CS:S:

```python
# ../players/teams.py

"""Provides player team functionality."""

# =============================================================================
# >> IMPORTS
# =============================================================================
# Site-Package Imports
#   ConfigObj
from configobj import ConfigObj

# Source.Python Imports
#   Core
from core import GAME_NAME
#   Paths
from paths import SP_DATA_PATH


# =============================================================================
# >> ALL DECLARATION
# =============================================================================
__all__ = ('classes_by_name',
           'classes_by_number',
           'team_managers',
           'teams_by_name',
           'teams_by_number',
           )


# =============================================================================
# >> GLOBAL VARIABLES
# =============================================================================
# Get the game's team data
_team_data = ConfigObj(
    SP_DATA_PATH / 'teams' / GAME_NAME + '.ini', unrepr=True)

# Store the game specific player teams by name
teams_by_name = _team_data.get('names', {'un': 0, 'spec': 1, 't': 2, 'ct': 3})

# Store the game specific player teams by number
teams_by_number = {number: alias for alias, number in teams_by_name.items()}

# Store the team manager names
team_managers = _team_data.get('manager', ())

# Make the managers an iterable if it is a string
if isinstance(team_managers, str):
    team_managers = [team_managers]

# Set the managers as a frozenset, since it should never be modified
team_managers = frozenset(team_managers)

# Store the game specific player classes by name
classes_by_name = _team_data.get('classes', {})

# Store the game specific player classes by number
classes_by_number = {number: name for name, number in classes_by_name.items()}

```

```
# ../data/source-python/teams/css.ini

manager = "cs_team_manager"

[names]
un = 0
spec = 1
t = 2
ct = 3


[classes]
Phoenix Connexion = 1
Elite Crew = 2
Artic Avengers = 3
Guerilla Warfare = 4
Seal Team 6 = 5
GSG-9 = 6
SAS = 7
GIGN = 8
```

```
from players.teams import classes_by_name
from players.teams import classes_by_number

class TestPlayer(Player):
    def get_team_name(self):
        """Return the name of the player's current team.

        :rtype: str
        """
        return teams_by_number.get(self.team, 'Unknown')

    def get_player_class_name(self):
        """Return the name of the player's current class.

        :rtype: int
        """
        return classes_by_number.get(self.player_class, 'Unknown')


print(classes_by_name)
print(classes_by_number)

pl = TestPlayer(1)
print(pl.get_team_name())
print(pl.get_player_class_name())
```

And here was the output:

```
{'Phoenix Connexion': 1, 'Elite Crew': 2, 'Artic Avengers': 3, 'Guerilla Warfare': 4, 'Seal Team 6': 5, 'GSG-9': 6, 'SAS': 7, 'GIGN': 8}
{1: 'Phoenix Connexion', 2: 'Elite Crew', 3: 'Artic Avengers', 4: 'Guerilla Warfare', 5: 'Seal Team 6', 6: 'GSG-9', 7: 'SAS', 8: 'GIGN'}
t
Phoenix Connexion
```

I used a subclass for testing purposes but of course this would be added into the Player class directly, as well as being bound to properties but you get the idea. I think it has the merit of being explicit as to whether to class/team are known for example prior to the players selecting any team/class when they joined the server or for games that are not supported. At least, for the later, this is more explicit than random `KeyError`.

PR #341, opened 2020-07-10
Fixed conversion to tuple from std::vector<DataType_t>.

CookStar:


jordanbriere:
Thanks!

For reference, this fixes:

```
TypeError: No to_python (by-value) converter found for C++ type: class std::vector<enum DataType_t,class std::allocator<enum DataType_t> >
```

PR #343, opened 2020-07-31
Properly updated DynamicHooks(libDynamicHooks.a).

CookStar:
Fix to d05527fd98a8dc734dc676361642ff9d4ff74d30.

Since libDynamicHooks.a is not recompiled, Segmentation Fault will be occurred at StackData.return_address.
https://github.com/Source-Python-Dev-Team/Source.Python/blob/7e06ce8e445e17425dfcecdad5895473c294fabc/src/core/modules/memory/memory_hooks.h#L65

If it is difficult to accept binary file due to security reasons, please edit the pull request, or close the pull request and commit the correct binary.


Ayuto:
Strange, I actually recompiled and copied the library. But apparently zero bytes have changed, so I must have grabbed the old version.

Thanks for reporting! Also thank you for providing the recompiled library! Though, I don't merge compiled binaries.

Nevertheless, I will keep this open as a reminder.

PR #344, opened 2020-08-05
Add default convention to CallingConvention.

CookStar:
This allows us to add a default calling convention when creating a custom calling convention.
In most cases, custom calling convention are derived from the basic calling convention, so this eliminates the need to re-implement the convention.

The initialization method is definitely weird, but I think it's impossible to implement any other way without breaking compatibility.

Also, this eliminates memory leak when the function is not hooked.
However, it does not eliminate the memory leak when the function is hooked.

Code example for a function that always ends with "ret 4".
(from https://forums.sourcepython.com/viewtopic.php?f=20&t=1528#p10195)

Before:
```python
# Source.Python Imports
#   Memory
from memory import get_data_type_size
from memory import CallingConvention
from memory import Convention
from memory import DataType
from memory import Register

class Retrun4(CallingConvention):
    def get_registers(self):
        registers = [Register.ESP]

        if (self.return_type == DataType.FLOAT or
            self.return_type == DataType.DOUBLE):
            registers.append(Register.ST0)
        else:
            registers.append(Register.EAX)
            if get_data_type_size(self.return_type) > 4:
                registers.append(Register.EDX);

        return registers

    def get_pop_size(self):
        return 4

    def get_argument_ptr(self, index, registers):
        offset = 4
        for i in range(index):
            offset += get_data_type_size(self.argument_types[i])
        return registers.esp.address.get_pointer() + offset

    def argument_ptr_changed(self, index, registers):
        pass

    def get_return_ptr(self, registers):
        if (self.return_type == DataType.FLOAT or
            self.return_type == DataType.DOUBLE):
            return registers.st0.address
        return registers.eax.address

    def return_ptr_changed(self, registers, return_ptr):
        pass
```

After:
```python
# Source.Python Imports
#   Memory
from memory import CallingConvention
from memory import Convention

class Retrun4(CallingConvention):
    def __init__(self, arg_types, return_type):
        super().__init__(arg_types, return_type,
                         default_convention=Convention.CDECL)

    def get_pop_size(self):
        return 4
```

Or:
```python
# Source.Python Imports
#   Memory
from memory import CallingConvention
from memory import Convention

class Retrun4(CallingConvention):
    default_convention = Convention.CDECL

    def get_pop_size(self):
        return 4
```


src/core/modules/memory/memory_function.cpp
```diff
@@ -188,6 +189,24 @@ CFunction::CFunction(unsigned long ulAddr, Convention_t eCallingConvention,
...
+
+			Py_DECREF(m_oCallingConvention.ptr());
+
+			delete _pCallingConventionWrapper;
+			m_pCallingConvention = NULL;
+		}
+
+		return;
+	}
+
```
Ayuto, line 209: I don't think this part is required. Boost should take care of deleting it.

src/core/modules/memory/memory_wrap.h
```diff
@@ -48,14 +56,55 @@ using namespace boost::python;
...
+			m_pCallingConvention = new x86GccCdecl(m_vecArgTypes, m_returnType, m_iAlignment);
+			break;
+		case CONV_THISCALL:
+			m_pCallingConvention = new x86GccThiscall(m_vecArgTypes, m_returnType, m_iAlignment);
+			break;
+		default:
+			BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Unsupported calling convention.")
+		}
+#endif
+	}
```
Ayuto, line 94: We could use ``MakeDynamicHooksConvention`` for this.

src/core/modules/memory/memory_function.cpp
CookStar: No, m_oCallingConvention will never be released, so it needs to Py_DECREF to release it when function is unhooked.

https://github.com/Source-Python-Dev-Team/Source.Python/blob/3d0d87502592180495c1327633752d381e5f3d61/src/core/modules/memory/memory_function.cpp#L158-L166

src/core/modules/memory/memory_wrap.h
CookStar: Fixed! https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/fb87d77359cac7e5b7b8fa32f5fca70b1e02e08f

src/core/modules/memory/memory_function.cpp
Ayuto: The Py_DECREF call is valid, but I actually meant the rest. ``m_oCallingConvention`` is created by Boost and when the reference counter reaches 0, Boost will take care of deleting its C++ instance.

CookStar: In my initial tests, Boost.Python did not call the destructor.

Also, in the test to verify the behavior, the destructor was not called if a pointer was passed to class_.

C++:
```c++
#include <string>
#include "export_main.h"

class Test
{
public:
	Test(){
		printf("Test\n");
	}

	~Test(){
		printf("~Test\n");
	}
};

void export_test(scope);

DECLARE_SP_MODULE(_test)
{
	export_test(_test);
}

void export_test(scope _test)
{
	class_<Test, Test *, boost::noncopyable>("Test", init<>())
	;
}
```

Python:
```python
from _test import Test

def test():
    print("make test")
    test_object = Test()
    del test_object
    print("deleted test")

test()
```

Output:
```
make test
Test
deleted test
```

I looked at the behavior of passing a pointer to class_ in the Boost.Python documentation and code, but I couldn't find a definitive answer.

I assumed that object's call policy would act like reference_existing_object when passing a pointer to class_, is this not the case?

Ayuto: I'm not fully sure why we added ``ICallingConventionWrapper *`` as the held type. I guess this was done to have full control over the deletion of the C++ instance.

I would remove the held type and make sure we decref only the instances that are not used by DynamicHooks.

CookStar: Fixed! https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/f2c403ca40225a5b1d47843a653ef083d043d415

CookStar: And bug fix. https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/025a42142073875e5d4a441c2a68c3c0a0f3beb1

Ayuto: Thank you! But it still looks like we have duplicated code in the destructor. I think would merge it with the second (already existing) part of the destructor as it's doing pretty much the same.

CookStar: That has been bothering me as well.
Anyway, Fixed! https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/f39a6c48a566987b573e8d4c049d307c69159268

CookStar: > I'm not fully sure why we added `ICallingConventionWrapper *` as the held type. I guess this was done to have full control over the deletion of the C++ instance.
> 
> I would remove the held type and make sure we decref only the instances that are not used by DynamicHooks.

I just remembered why `ICallingConventionWrapper *` was added.
When Source.Python unloads, destructor is called and m_pCallingConvention will be freed, but if the function was hooked, DynamicHooks also frees m_pCallingConvention, therefore it will double-free and crash.

CookStar: I don't think there is an easy solution other than the original method(https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/adbc65a9324276aeb863e421b14e001d0514e476).
How do you think?

jordanbriere: > When Source.Python unloads, destructor is called and m_pCallingConvention will be freed, but if the function was hooked, DynamicHooks also frees m_pCallingConvention, therefore it will double-free and crash.

Which will most likely still happens because you removed that check: [../memory_function.cpp#L197-L199](https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/files#diff-200b956ad6eb0205bdea4cfc3a9aea18L197-L199). If it doesn't, then the built-in conventions created [here](https://github.com/Source-Python-Dev-Team/Source.Python/blob/adbc65a9324276aeb863e421b14e001d0514e476/src/core/modules/memory/memory_function.cpp#L145) are probably leaking again. 🤔

CookStar: > Which will most likely still happens because you removed that check: [../memory_function.cpp#L197-L199](https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/files#diff-200b956ad6eb0205bdea4cfc3a9aea18L197-L199).

When the function is hooked, `m_bAllocatedCallingConvention` is set to false, so it should not be double-free.
https://github.com/Source-Python-Dev-Team/Source.Python/blob/f39a6c48a566987b573e8d4c049d307c69159268/src/core/modules/memory/memory_function.cpp#L383-L388
The same goes for built-in conventions.

jordanbriere: > When the function is hooked, m_bAllocatedCallingConvention is set to false, so it should not be double-free.

Yes, that's right. I forgot I added that. To be honest, trying to guess who own and who will take care of the convention is rather a nightmare. Currently, the ones that are allocated by `CFunction` are taken care of, but the ones that are instantiated by Boost from Python are leaking. The easiest solution would be for DynamicHooks to accept a `shared_ptr` rather than a raw pointer, then everyone could get along and be happy.

CookStar: I certainly think that's the most complete way to do it, but I personally don't think it should be done with this Pull Request.

CookStar: > The easiest solution would be for DynamicHooks to accept a `shared_ptr` rather than a raw pointer, then everyone could get along and be happy.

Something like this?
https://github.com/CookStar/DynamicHooks/commit/ed7fc6148db3dfeb6da210f2806907f44eef9866
https://github.com/CookStar/Source.Python/commit/2a04d2f72532d4c4ea982c46e10ba6b80d6f1218

jordanbriere: > Something like this?

Something like that, yes. Along with the `CallingConvention` python class to be held and initialized as shared. Not sure if that would be the right implementation though, as setting `-std=c++11` is likely to bump our prerequisites, no? But yeah, basically, so long as the Python class, `CFunction`, and `CHook` all use a shared pointer they can free whenever they are done with it without affecting the others, etc. it would be ideal.

CookStar: > Along with the `CallingConvention` python class to be held and initialized as shared.

 If we wrapped shared_ptr in python class, I don't think we would be able to Py_DECREF because the python class would not release the pointer.

CookStar: > Not sure if that would be the right implementation though, as setting `-std=c++11` is likely to bump our prerequisites, no?

You're right, and I honestly don't think it's my place to judge.

CookStar: Ok, surely removing this [../memory_function.cpp#L197-L199](https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/files#diff-200b956ad6eb0205bdea4cfc3a9aea18L197-L199) may be problematic. https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/854d96fd4f7d539782db15fdf819c8a902a96103

CookStar: We may want to add C++ interface to CFunction.

CookStar: > We may want to add C++ interface to CFunction.

Done. https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/0062175b7cc5147d69470a414c61a978c1557995

CookStar: > The easiest solution would be for DynamicHooks to accept a `shared_ptr` rather than a raw pointer, then everyone could get along and be happy.

> Something like this?
> [CookStar/DynamicHooks@ed7fc61](https://github.com/CookStar/DynamicHooks/commit/ed7fc6148db3dfeb6da210f2806907f44eef9866)
> [CookStar@2a04d2f](https://github.com/CookStar/Source.Python/commit/2a04d2f72532d4c4ea982c46e10ba6b80d6f1218)
> 
> Edit:
> Little corrections. [CookStar@3254ba0](https://github.com/CookStar/Source.Python/commit/3254ba0148a739110468763364e6500322f4af4a)

@Ayuto I'd like to know your opinion about shared_ptr.

CookStar:
Resolved conflicts. https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/3c4935b0b865245976142924455c52015b9f13d9

src/core/modules/memory/memory_function.cpp
jordanbriere: https://github.com/Ayuto/DynamicHooks/pull/7 and https://github.com/CookStar/Source.Python/pull/1 appears to fix all leaks/issues related to the convention from my testings.

CookStar: Ok, first of all there are two problems with this implementation.

First, since m_oCallingConvention is tied to CFunction, m_oCallingConvention is also destroyed with the destruction of CFunction. In principle, m_oCallingConvention should remain until the hook is deleted.

Second, the Python override will not succeed in this implementation. As you can see in the Output, `get_pop_size` is not called.

Code:
```python
from memory import CallingConvention
from memory import Convention
from memory import DataType
from memory.hooks import PostHook
from memory.hooks import PreHook

server = find_binary("server", srv_check=False)

class Return4(CallingConvention):
    def __init__(self, arg_types, return_type):
        super().__init__(arg_types, return_type,
                         default_convention=Convention.CDECL)
        print("__init__")
    def __del__(self):
        print("__del__")
    def get_pop_size(self):
        print("get_pop_size")
        return 4

def test():
    CCSPlayer_Weapon_ShootPosition = server[b"\x55\x89\xE5\x83\xEC\x18\x89\x5D\xF8\x8B\x5D\x08\x89\x75\xFC\x8B\x75\x0C\x89\x1C\x24\x89\x74\x24\x04\xE8\x2A\x2A\x2A\x2A\x83\xEC\x04"].make_function(
        Return4, (
            DataType.POINTER,
            DataType.POINTER,
        ),
        DataType.VOID,
    )

    @PreHook(CCSPlayer_Weapon_ShootPosition)
    def pre_CCSPlayer_Weapon_ShootPosition(args):
        pass

    @PostHook(CCSPlayer_Weapon_ShootPosition)
    def post_CCSPlayer_Weapon_ShootPosition(args, return_value):
        pass

test()
```
Output:
```
sp plugin load test
__init__
[SP] Successfully loaded plugin 'test'.
sp plugin unload test
[SP] Unloading plugin 'test'...
__del__
[SP] Successfully unloaded plugin 'test'.
```

jordanbriere: Arghh. I guess I was too focus about getting them deleted to notice. 😆 

How does replacing [this line](https://github.com/CookStar/Source.Python/blob/205f776507a889fce4cf7274529ed0368704f59e/src/core/modules/memory/memory_function.cpp#L158) to:

```cpp
		m_pCallingConvention = extract<ICallingConvention*>(m_oCallingConvention);
		Py_INCREF(m_oCallingConvention.ptr());
		detail::initialize_wrapper(m_oCallingConvention.ptr(), get_pointer((ICallingConventionWrapper *)m_pCallingConvention));
```

And [these two lines](https://github.com/CookStar/Source.Python/blob/205f776507a889fce4cf7274529ed0368704f59e/src/core/modules/memory/memory_function.cpp#L181-L182) to:

```cpp
	if (m_pCallingConvention && !m_pCallingConvention->m_bHooked)
	{
		if (m_oCallingConvention.is_none())
			delete m_pCallingConvention;
		else
			Py_DECREF(m_oCallingConvention.ptr());
	}
```

Works for you?

<details>
 <summary>Full file ←</summary>

```cpp
/**
* =============================================================================
* Source Python
* Copyright (C) 2012-2015 Source Python Development Team.  All rights reserved.
* =============================================================================
*
* This program is free software; you can redistribute it and/or modify it under
* the terms of the GNU General Public License, version 3.0, as published by the
* Free Software Foundation.
*
* This program is distributed in the hope that it will be useful, but WITHOUT
* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
* FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
* details.
*
* You should have received a copy of the GNU General Public License along with
* this program.  If not, see <http://www.gnu.org/licenses/>.
*
* As a special exception, the Source Python Team gives you permission
* to link the code of this program (as well as its derivative works) to
* "Half-Life 2," the "Source Engine," and any Game MODs that run on software
* by the Valve Corporation.  You must obey the GNU General Public License in
* all respects for all other code used.  Additionally, the Source.Python
* Development Team grants this exception to all derivative works.
*/

// ============================================================================
// >> INCLUDES
// ============================================================================
// DynCall
#include "dyncall.h"
#include "dyncall_signature.h"

// Memory
#include "memory_function.h"
#include "memory_utilities.h"
#include "memory_hooks.h"
#include "memory_wrap.h"

// DynamicHooks
#include "conventions/x86MsCdecl.h"
#include "conventions/x86MsThiscall.h"
#include "conventions/x86MsStdcall.h"
#include "conventions/x86MsFastcall.h"
#include "conventions/x86GccCdecl.h"
#include "conventions/x86GccThiscall.h"

// Source.Python
#include "utilities/call_python.h"


// ============================================================================
// >> EXTERNALS
// ============================================================================
extern std::map<CHook *, std::map<HookType_t, std::list<object> > > g_mapCallbacks;


// ============================================================================
// >> GLOBAL VARIABLES
// ============================================================================
DCCallVM* g_pCallVM = dcNewCallVM(4096);


// ============================================================================
// >> GetDynCallConvention
// ============================================================================
int GetDynCallConvention(Convention_t eConv)
{
	switch (eConv)
	{
		case CONV_CUSTOM: return -1;
		case CONV_CDECL: return DC_CALL_C_DEFAULT;
		case CONV_THISCALL:
			#ifdef _WIN32
				return DC_CALL_C_X86_WIN32_THIS_MS;
			#else
				return DC_CALL_C_X86_WIN32_THIS_GNU;
			#endif
#ifdef _WIN32
		case CONV_STDCALL: return DC_CALL_C_X86_WIN32_STD;
		case CONV_FASTCALL: return DC_CALL_C_X86_WIN32_FAST_MS;
#endif
	}

	BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Unsupported calling convention.")
	return -1;
}


// ============================================================================
// >> MakeDynamicHooksConvention
// ============================================================================
ICallingConvention* MakeDynamicHooksConvention(Convention_t eConv, std::vector<DataType_t> vecArgTypes, DataType_t returnType, int iAlignment)
{
#ifdef _WIN32
	switch (eConv)
	{
	case CONV_CDECL: return new x86MsCdecl(vecArgTypes, returnType, iAlignment);
	case CONV_THISCALL: return new x86MsThiscall(vecArgTypes, returnType, iAlignment);
	case CONV_STDCALL: return new x86MsStdcall(vecArgTypes, returnType, iAlignment);
	case CONV_FASTCALL: return new x86MsFastcall(vecArgTypes, returnType, iAlignment);
	}
#else
	switch (eConv)
	{
	case CONV_CDECL: return new x86GccCdecl(vecArgTypes, returnType, iAlignment);
	case CONV_THISCALL: return new x86GccThiscall(vecArgTypes, returnType, iAlignment);
	}
#endif

	BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Unsupported calling convention.")
	return NULL;
}


// ============================================================================
// >> CFunction
// ============================================================================
CFunction::CFunction(unsigned long ulAddr, object oCallingConvention, object oArgs, object oReturnType)
	:CPointer(ulAddr)
{
	// Step 1: Validate and convert the argument types
	m_tArgs = tuple(oArgs);

	// Step 2: Determine the return type/converter
	try
	{
		// If this line succeds...
		m_eReturnType = extract<DataType_t>(oReturnType);

		// ...no converter will be used
		m_oConverter = object();
	}
	catch( ... )
	{
		PyErr_Clear();

		// If this happens the return type is a converter for a pointer
		m_eReturnType = DATA_TYPE_POINTER;
		m_oConverter = oReturnType;
	}

	// Step 3: Determine the calling convention
	try
	{
		// If this line succeeds the user wants to create a function with the built-in calling conventions
		m_eCallingConvention = extract<Convention_t>(oCallingConvention);
		m_pCallingConvention = MakeDynamicHooksConvention(m_eCallingConvention, ObjectToDataTypeVector(m_tArgs), m_eReturnType);
		m_oCallingConvention = object();
	}
	catch( ... )
	{
		PyErr_Clear();

		// A custom calling convention will be used...
		m_eCallingConvention = CONV_CUSTOM;
		m_oCallingConvention = oCallingConvention(m_tArgs, m_eReturnType);
		m_pCallingConvention = extract<ICallingConvention*>(m_oCallingConvention);
		Py_INCREF(m_oCallingConvention.ptr());
		detail::initialize_wrapper(m_oCallingConvention.ptr(), get_pointer((ICallingConventionWrapper *)m_pCallingConvention));
	}

	// Step 4: Get the DynCall calling convention
	m_iCallingConvention = GetDynCallConvention(m_eCallingConvention);
}

CFunction::CFunction(unsigned long ulAddr, Convention_t eCallingConvention,
	int iCallingConvention, tuple tArgs, DataType_t eReturnType, object oConverter)
	:CPointer(ulAddr)
{
	m_eCallingConvention = eCallingConvention;
	m_iCallingConvention = iCallingConvention;
	m_pCallingConvention = NULL;
	m_oCallingConvention = object();

	m_tArgs = tArgs;
	m_eReturnType = eReturnType;
	m_oConverter = oConverter;
}

CFunction::~CFunction()
{
	if (m_pCallingConvention && !m_pCallingConvention->m_bHooked)
	{
		if (m_oCallingConvention.is_none())
			delete m_pCallingConvention;
		else
			Py_DECREF(m_oCallingConvention.ptr());
	}

	m_pCallingConvention = NULL;
}

bool CFunction::IsCallable()
{
	return (m_eCallingConvention != CONV_CUSTOM) && (m_iCallingConvention != -1);
}

bool CFunction::IsHookable()
{
	return m_pCallingConvention != NULL;
}

bool CFunction::IsHooked()
{
	return GetHookManager()->FindHook((void *) m_ulAddr) != NULL;
}

CFunction* CFunction::GetTrampoline()
{
	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
	if (!pHook)
		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function was not hooked.")

	return new CFunction((unsigned long) pHook->m_pTrampoline, m_eCallingConvention,
		m_iCallingConvention, m_tArgs, m_eReturnType, m_oConverter);
}

template<class ReturnType, class Function>
ReturnType CallHelper(Function func, DCCallVM* vm, unsigned long addr)
{
	ReturnType result;
	TRY_SEGV()
		result = (ReturnType) func(vm, addr);
	EXCEPT_SEGV()
	return result;
}

void CallHelperVoid(DCCallVM* vm, unsigned long addr)
{
	TRY_SEGV()
		dcCallVoid(vm, addr);
	EXCEPT_SEGV()
}

object CFunction::Call(tuple args, dict kw)
{
	if (!IsCallable())
		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function is not callable.")

	Validate();
	if (len(args) != len(m_tArgs))
		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Number of passed arguments is not equal to the required number.")

	// Reset VM and set the calling convention
	dcReset(g_pCallVM);
	dcMode(g_pCallVM, m_iCallingConvention);

	// Loop through all passed arguments and add them to the VM
	for(int i=0; i < len(args); i++)
	{
		object arg = args[i];
		switch(extract<DataType_t>(m_tArgs[i]))
		{
			case DATA_TYPE_BOOL:		dcArgBool(g_pCallVM, extract<bool>(arg)); break;
			case DATA_TYPE_CHAR:		dcArgChar(g_pCallVM, extract<char>(arg)); break;
			case DATA_TYPE_UCHAR:		dcArgChar(g_pCallVM, extract<unsigned char>(arg)); break;
			case DATA_TYPE_SHORT:		dcArgShort(g_pCallVM, extract<short>(arg)); break;
			case DATA_TYPE_USHORT:		dcArgShort(g_pCallVM, extract<unsigned short>(arg)); break;
			case DATA_TYPE_INT:			dcArgInt(g_pCallVM, extract<int>(arg)); break;
			case DATA_TYPE_UINT:		dcArgInt(g_pCallVM, extract<unsigned int>(arg)); break;
			case DATA_TYPE_LONG:		dcArgLong(g_pCallVM, extract<long>(arg)); break;
			case DATA_TYPE_ULONG:		dcArgLong(g_pCallVM, extract<unsigned long>(arg)); break;
			case DATA_TYPE_LONG_LONG:	dcArgLongLong(g_pCallVM, extract<long long>(arg)); break;
			case DATA_TYPE_ULONG_LONG:	dcArgLongLong(g_pCallVM, extract<unsigned long long>(arg)); break;
			case DATA_TYPE_FLOAT:		dcArgFloat(g_pCallVM, extract<float>(arg)); break;
			case DATA_TYPE_DOUBLE:		dcArgDouble(g_pCallVM, extract<double>(arg)); break;
			case DATA_TYPE_POINTER:
			{
				unsigned long ulAddr = 0;
				if (arg.ptr() != Py_None)
					ulAddr = ExtractPointer(arg)->m_ulAddr;

				dcArgPointer(g_pCallVM, ulAddr);
				break;
			}
			case DATA_TYPE_STRING:		dcArgPointer(g_pCallVM, (unsigned long) (void *) extract<char *>(arg)); break;
			default:					BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Unknown argument type.")
		}
	}

	// Call the function
	switch(m_eReturnType)
	{
		case DATA_TYPE_VOID:		CallHelperVoid(g_pCallVM, m_ulAddr); break;
		case DATA_TYPE_BOOL:		return object(CallHelper<bool>(dcCallBool, g_pCallVM, m_ulAddr));
		case DATA_TYPE_CHAR:		return object(CallHelper<char>(dcCallChar, g_pCallVM, m_ulAddr));
		case DATA_TYPE_UCHAR:		return object(CallHelper<unsigned char>(dcCallChar, g_pCallVM, m_ulAddr));
		case DATA_TYPE_SHORT:		return object(CallHelper<short>(dcCallShort, g_pCallVM, m_ulAddr));
		case DATA_TYPE_USHORT:		return object(CallHelper<unsigned short>(dcCallShort, g_pCallVM, m_ulAddr));
		case DATA_TYPE_INT:			return object(CallHelper<int>(dcCallInt, g_pCallVM, m_ulAddr));
		case DATA_TYPE_UINT:		return object(CallHelper<unsigned int>(dcCallInt, g_pCallVM, m_ulAddr));
		case DATA_TYPE_LONG:		return object(CallHelper<long>(dcCallLong, g_pCallVM, m_ulAddr));
		case DATA_TYPE_ULONG:		return object(CallHelper<unsigned long>(dcCallLong, g_pCallVM, m_ulAddr));
		case DATA_TYPE_LONG_LONG:	return object(CallHelper<long long>(dcCallLongLong, g_pCallVM, m_ulAddr));
		case DATA_TYPE_ULONG_LONG:	return object(CallHelper<unsigned long long>(dcCallLongLong, g_pCallVM, m_ulAddr));
		case DATA_TYPE_FLOAT:		return object(CallHelper<float>(dcCallFloat, g_pCallVM, m_ulAddr));
		case DATA_TYPE_DOUBLE:		return object(CallHelper<double>(dcCallDouble, g_pCallVM, m_ulAddr));
		case DATA_TYPE_POINTER:
		{
			CPointer pPtr = CPointer(CallHelper<unsigned long>(dcCallPointer, g_pCallVM, m_ulAddr));
			if (!m_oConverter.is_none())
				return m_oConverter(pPtr);

			return object(pPtr);
		}
		case DATA_TYPE_STRING:		return object(CallHelper<const char *>(dcCallPointer, g_pCallVM, m_ulAddr));
		default:					BOOST_RAISE_EXCEPTION(PyExc_TypeError, "Unknown return type.")
	}
	return object();
}

object CFunction::CallTrampoline(tuple args, dict kw)
{
	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
	if (!pHook)
		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function was not hooked.")

	return CFunction((unsigned long) pHook->m_pTrampoline, m_eCallingConvention,
		m_iCallingConvention, m_tArgs, m_eReturnType, m_oConverter).Call(args, kw);
}

object CFunction::SkipHooks(tuple args, dict kw)
{
	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
	if (pHook)
		return CFunction((unsigned long) pHook->m_pTrampoline, m_eCallingConvention,
			m_iCallingConvention, m_tArgs, m_eReturnType, m_oConverter).Call(args, kw);

	return Call(args, kw);
}

CHook* HookFunctionHelper(void* addr, ICallingConvention* pConv)
{
	CHook* result;
	TRY_SEGV()
		result = GetHookManager()->HookFunction(addr, pConv);
	EXCEPT_SEGV()
	return result;
}

void CFunction::AddHook(HookType_t eType, PyObject* pCallable)
{
	if (!IsHookable())
		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function is not hookable.")

	Validate();
	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);

	// Prepare arguments for log message
	str type = str(eType);
	const char* szType = extract<const char*>(type);

	str convention = str(m_eCallingConvention);
	const char* szConvention = extract<const char*>(convention);

	str args = str(m_tArgs);
	const char* szArgs = extract<const char*>(args);

	str return_type = str(m_eReturnType);
	const char* szReturnType = extract<const char*>(return_type);

	object oCallback = object(handle<>(borrowed(pCallable)));
	str callback = str(oCallback);
	const char* szCallback = extract<const char*>(callback);

	PythonLog(
		4,
		"Hooking function: type=%s, addr=%u, conv=%s, args=%s, rtype=%s, callback=%s",
		szType,
		m_ulAddr,
		szConvention,
		szArgs,
		szReturnType,
		szCallback
	);

	if (!pHook) {
		pHook = HookFunctionHelper((void *) m_ulAddr, m_pCallingConvention);
	}

	// Add the hook handler. If it's already added, it won't be added twice
	pHook->AddCallback(eType, (HookHandlerFn *) (void *) &SP_HookHandler);
	g_mapCallbacks[pHook][eType].push_back(object(handle<>(borrowed(pCallable))));
}

bool CFunction::AddHook(HookType_t eType, HookHandlerFn* pFunc)
{
	if (!IsHookable())
		return false;

	CHook* pHook = GetHookManager()->FindHook((void*) m_ulAddr);

	if (!pHook) {
		pHook = GetHookManager()->HookFunction((void*) m_ulAddr, m_pCallingConvention);

		if (!pHook)
			return false;
	}

	pHook->AddCallback(eType, pFunc);
	return true;
}

void CFunction::RemoveHook(HookType_t eType, PyObject* pCallable)
{
	Validate();
	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
	if (!pHook)
		return;

	g_mapCallbacks[pHook][eType].remove(object(handle<>(borrowed(pCallable))));
}

void CFunction::DeleteHook()
{
	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
	if (!pHook)
		return;

	g_mapCallbacks.erase(pHook);
	// Set the calling convention to NULL, because DynamicHooks will delete it otherwise.
	pHook->m_pCallingConvention->m_bHooked = false;
	pHook->m_pCallingConvention = NULL;
	GetHookManager()->UnhookFunction((void *) m_ulAddr);
}
```

</details>

CookStar: Now the Python object is leaking again.

Output:
```
sp plugin load test
[SP] Loading plugin 'test'...
__init__
get_pop_size
get_pop_size
[SP] Successfully loaded plugin 'test'.
sp plugin reload test
[SP] Unloading plugin 'test'...
[SP] Successfully unloaded plugin 'test'.
[SP] Loading plugin 'test'...
__init__
[SP] Successfully loaded plugin 'test'.
sp plugin unload test
[SP] Unloading plugin 'test'...
__del__
[SP] Successfully unloaded plugin 'test'
```

This solution will not solve the problem.
As long as m_oCallingConvention's resources are managed by CFunction, the leak cannot be resolved. Because we have already passed the ownership of m_oCallingConvention to CHook.

CookStar: Now the Python object is leaking again.
This code is functionally identical to my pull requested code here.

Code2(test2.py):
```python
from memory import CallingConvention
from memory import Convention
from memory import DataType
from memory.hooks import PostHook
from memory.hooks import PreHook

server = find_binary("server", srv_check=False)

class Return4(CallingConvention):
    def __init__(self, arg_types, return_type):
        super().__init__(arg_types, return_type,
                         default_convention=Convention.CDECL)
        print("__init__", id(self))
    def __del__(self):
        print("__del__", id(self))
    def get_pop_size(self):
        print("get_pop_size")
        return 4

def test2():
    CCSPlayer_Weapon_ShootPosition = server[b"\x55\x89\xE5\x83\xEC\x18\x89\x5D\xF8\x8B\x5D\x08\x89\x75\xFC\x8B\x75\x0C\x89\x1C\x24\x89\x74\x24\x04\xE8\x2A\x2A\x2A\x2A\x83\xEC\x04"].make_function(
        Return4, (
            DataType.POINTER,
            DataType.POINTER,
        ),
        DataType.VOID,
    )

    @PreHook(CCSPlayer_Weapon_ShootPosition)
    def pre_CCSPlayer_Weapon_ShootPosition(args):
        pass

    @PostHook(CCSPlayer_Weapon_ShootPosition)
    def post_CCSPlayer_Weapon_ShootPosition(args, return_value):
        pass

test2()
```
Code3(test3.py):
```python
from memory import CallingConvention
from memory import Convention
from memory import DataType

server = find_binary("server", srv_check=False)

class Return4(CallingConvention):
    def __init__(self, arg_types, return_type):
        super().__init__(arg_types, return_type,
                         default_convention=Convention.CDECL)
        print("__init__", id(self))
    def __del__(self):
        print("__del__", id(self))
    def get_pop_size(self):
        print("get_pop_size")
        return 4

def test3():
    CCSPlayer_Weapon_ShootPosition = server[b"\x55\x89\xE5\x83\xEC\x18\x89\x5D\xF8\x8B\x5D\x08\x89\x75\xFC\x8B\x75\x0C\x89\x1C\x24\x89\x74\x24\x04\xE8\x2A\x2A\x2A\x2A\x83\xEC\x04"].make_function(
        Return4, (
            DataType.POINTER,
            DataType.POINTER,
        ),
        DataType.VOID,
    )

    CCSPlayer_Weapon_ShootPosition._delete_hook()

test3()
```

Output:

```
sp plugin load test2
[SP] Loading plugin 'test2'...
__init__ 3978578336
get_pop_size
get_pop_size
[SP] Successfully loaded plugin 'test2'.
sp plugin unload test2
[SP] Unloading plugin 'test2'...
[SP] Successfully unloaded plugin 'test2'.
sp plugin load test3
[SP] Loading plugin 'test3'...
__init__ 3978711744
__del__ 3978711744
[SP] Successfully loaded plugin 'test3'.
sp plugin unload test3
[SP] Unloading plugin 'test3'...
[SP] Successfully unloaded plugin 'test3'.
```

This solution will not solve the problem.
As long as m_oCallingConvention's resources are managed by CFunction, the leak cannot be resolved. Because we have already passed the ownership of m_oCallingConvention to CHook.

jordanbriere: > Your code is functionally identical to my pull requested code here.

It fixes at least two other leaks this PR is introducing or still suffering. First, it fixes the held type of `CallingConvention`, so that unhooked conventions that are not managed by `CFunction` are properly taken care of and then it fixes `m_pDefaultCallingConvention` never being deleted because we (and `DynamicHooks`) delete the `ICallingConvention` pointers, not `ICallingConventionWrapper`. Therefore, we never call the derived destructor; hence why it needs to be virtual.

> This solution will not solve the problem.
> As long as m_oCallingConvention's resources are managed by CFunction, the leak cannot be resolved. Because we have a
> lready passed the ownership of m_oCallingConvention to CHook.

To be honest, I wasn't too concerned about hooked conventions because functions should ideally never be unhooked on the fly so that the chain is respected and we don't clash with other plugins such as `SourceMod`, etc. that may still fallback to our handler (hence why `Function._delete_hook` is private).

However, this also means [memory.Callback](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/memory/__init__.py#L136)'s conventions are currently leaking. Though, this is really not that hard to fix. All we have to do is release the strong references we reserved for `DynamicHooks` when we manually delete the hook and want to keep it's convention alive (in case the `Function` that allocated it, or anyone else referencing it, is still around): 

```cpp
void CFunction::DeleteHook()
{
	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
	if (!pHook)
		return;

	g_mapCallbacks.erase(pHook);

	// Release the Python reference we reserved for DynamicHooks.
	ICallingConventionWrapper *pConv = dynamic_cast<ICallingConventionWrapper *>(pHook->m_pCallingConvention);
	if (pConv)
	{
		PyObject *pOwner = detail::wrapper_base_::owner(pConv);
		if (pOwner && Py_REFCNT(pOwner))
			Py_DECREF(pOwner);
	}
	// Flag the convention as no longer hooked and being taken care of by DynamicHooks.
	pHook->m_pCallingConvention->m_bHooked = false;

	// Set the calling convention to NULL, because DynamicHooks will delete it otherwise.
	pHook->m_pCallingConvention = NULL;
	GetHookManager()->UnhookFunction((void *) m_ulAddr);
}
```

And when `DynamicHooks` deletes them [when we unload](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/sp_main.cpp#L375):

```cpp
	virtual ~ICallingConventionWrapper()
	{
		// If we are still flagged as hooked, then that means DynamicHooks is done with us.
		if (m_bHooked)
		{
			// Release the Python reference we reserved for DynamicHooks.
			PyObject *pOwner = detail::wrapper_base_::owner(this);
			if (pOwner && Py_REFCNT(pOwner))
				Py_DECREF(pOwner);
		}

		delete m_pDefaultCallingConvention;
		m_pDefaultCallingConvention = nullptr;
	}
```

I've re-forked your branch and the full updated code can be found [here](https://github.com/jordanbriere/Source.Python/commit/8aab63b1b9bfafe361f1e5f4cbae1a36d9ab302d).

CookStar: > It fixes at least two other leaks this PR is introducing or still suffering. First, it fixes the held type of `CallingConvention`, so that unhooked conventions that are not managed by `CFunction` are properly taken care of and then it fixes `m_pDefaultCallingConvention` never being deleted for hooked conventions because `DynamicHooks` delete the `ICallingConvention` pointers, not `ICallingConventionWrapper`. Therefore, it never call the derived destructor, which is why it needs to be virtual.

You are correct! The leak was not only during `CHookManager::UnhookAllFunctions`, but the default calling convention destructor was not called either, so it seems to have been leaking(`x86MsCdecl`, etc.).

I believe this fixes all the leaks except for leak when you combine add_hook and _delete_hook. It's a niche case and I don't think there is a problem.
And I didn't have any problems with the test.
Thanks a lot!

jordanbriere: > You are correct! The leak was not only during `CHookManager::UnhookAllFunctions`, but the default calling convention destructor was not called either, so it seems to have been leaking(`x86MsCdecl`, etc.).

Good point! Their buffer were more than certainly leaking.

> I believe this fixes all the leaks except for leak when you combine add_hook and _delete_hook. It's a niche case and I don't think there is a problem.

Got a code that reproduce? I ran a quick tests just now and re-hooked a function over 100 million times and the memory didn't grow.

> And I didn't have any problems with the test.
> Thanks a lot!

Awesome! Everything seems to be working fine on my side as well. The only scenario I know would cause issues, is if someone externally store the python instance passed to an override, manually delete the hook and then try to interact with the convention they stored. The pointer wrapped by the proxy object would be freed at that point, resulting into undefined behaviours (likely a segfault). But yeah, if you are dumb enough to do that then you probably deserve everything coming your way. :trollface: 

Something else I would like to fix, is the wrapper only being initialized by conventions managed by `CFunction`. It seems to be an implementation detail missing when using a custom constructor. Whether this is an oversight or intended design I'm not too sure. I brainstormed a few generic ideas that would address such case going forward but haven't had a chance to put anything in practice yet.

Another point I think we should find an alternative is as you mentioned above:

> The initialization method is definitely weird, but I think it's impossible to implement any other way without breaking compatibility.

Having to override `__init__` to set a default convention is rather ugly. Perhaps something like this could be preferable:

```python
class Retrun4(CallingConvention):
    default_convention = Convention.CDECL

    def get_pop_size(self):
        return 4
```

I guess it could easily be done by default `m_pDefaultConvention` to `NULL` in the constructor, then declaring an accessor that resolve it once from the owner. Something like that:

```cpp
ICallingConvention *GetDefaultConvention()
{
    if (m_pDefaultCallingConvention)
        return m_pDefaultCallingConvention;

    // resolve default_convention from the owner
    // assign it to m_pDefaultCallingConvention

    return m_pDefaultCallingConvention;
}
```

The usage would certainly feel better and more natural to me. What you think?

CookStar: > Got a code that reproduce? I ran a quick tests just now and re-hooked a function over 100 million times and the memory didn't grow.

Code:
```python
from memory import CallingConvention
from memory import Convention
from memory import DataType
from memory.hooks import HookType

server = find_binary("server", srv_check=False)

#Note, a function for testing.(Not CDECL.)
pointer = server[b"\x55\x89\xE5\x83\xEC\x18\x89\x5D\xF8\x8B\x5D\x08\x89\x75\xFC\x8B\x75\x0C\x89\x1C\x24\x89\x74\x24\x04\xE8\x2A\x2A\x2A\x2A\x83\xEC\x04"]

def make_hook():
    CCSPlayer_Weapon_ShootPosition = pointer.make_function(
        Convention.CDECL, (
            DataType.POINTER,
            DataType.POINTER,
        ),
        DataType.VOID,
    )

    def pre_CCSPlayer_Weapon_ShootPosition(args):
        pass

    CCSPlayer_Weapon_ShootPosition.add_hook(HookType.PRE, pre_CCSPlayer_Weapon_ShootPosition)

def delete_hook():
    CCSPlayer_Weapon_ShootPosition = pointer.make_function(
        Convention.CDECL, (
            DataType.POINTER,
            DataType.POINTER,
        ),
        DataType.VOID,
    )

    CCSPlayer_Weapon_ShootPosition._delete_hook()

make_hook()
delete_hook()
```

When the function hooked with the normal calling convention has already been released, calling _delete_hook from another function will leak m_pCallingConvention. However, there is no guarantee the function has been released, so you can't delete it in _delete_hook, and I don't think it's a problem to leave it alone.


> Another point I think we should find an alternative is as you mentioned above:
> 
> > The initialization method is definitely weird, but I think it's impossible to implement any other way without breaking compatibility.
> 
> Having to override `__init__` to set a default convention is rather ugly. Perhaps something like this could be preferable:
> 
> ```python
> class Retrun4(CallingConvention):
>     default_convention = Convention.CDECL
> 
>     def get_pop_size(self):
>         return 4
> ```

I think this is a good approach. How about this kind of implementation? https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/b4cb2c02b10139ac1a0c90b660a5f17c09c07d70


> I guess it could easily be done by default `m_pDefaultConvention` to `NULL` in the constructor, then declaring an accessor that resolve it once from the owner. Something like that:
> 
> ```c++
> ICallingConvention *GetDefaultConvention()
> {
>     if (m_pDefaultCallingConvention)
>         return m_pDefaultCallingConvention;
> 
>     // resolve default_convention from the owner
>     // assign it to m_pDefaultCallingConvention
> 
>     return m_pDefaultCallingConvention;
> }
> ```
> 
> The usage would certainly feel better and more natural to me. What you think?

Since `m_pDefaultCallingConvention` is currently used to check for overrides, it's not a good idea to not return `NULL`. Also, I don't think it's necessary to check the python attribute every time when it returns `NULL`.

CookStar: > Something else I would like to fix, is the wrapper only being initialized by conventions managed by `CFunction`. It seems to be an implementation detail missing when using a custom constructor. Whether this is an oversight or intended design I'm not too sure. I brainstormed a few generic ideas that would address such case going forward but haven't had a chance to put anything in practice yet.

I couldn't understand the problem. Is this about general documentation? Or is it about the docstrings?

jordanbriere: > When the function hooked with the normal calling convention has already been released, calling `_delete_hook` from another function will leak `m_pCallingConvention`. However, there is no guarantee the function has been released, so you can't delete it in `_delete_hook`, and I don't think it's a problem to leave it alone.

That's right, if `m_pCallingConvention` isn't a `ICallingConventionWrapper`, it won't be freed ever in such case. I agree though, very specific scenario that is unlikely to happens.

> I couldn't understand the problem. Is this about general documentation? Or is it about the docstring?

The problem isn't about documentation, but that `CallingConvention` should be self-initializing and shouldn't rely on `CFunction` to be fully functional. The following addresses that: [282bf03](https://github.com/CookStar/Source.Python/compare/add_default_callingconvention...jordanbriere:add_default_callingconvention)

It also change the retrieval priority of `default_convention` which should be a fallback if no default convention was passed on construction of the instance itself. Basically, this:

```python
class Conv(CallingConvention):
    default_convention = Convention.THISCALL

conv = Conv(..., default_convention=Convention.CDECL)
```

Should prioritize `CDECL` because it was explicitly given on instantiation of the instance, as opposed to be set to the class itself so it follows Python resolution order policies.

CookStar: > **EDIT:** I was thinking about that while doing something else and we can actually [fix it with a bit of ingenuity](https://github.com/jordanbriere/Source.Python/commit/6d26bca14429b498e5862ccbbfbae57fd0bb7c8a).

I have a little question, why are you checking `ICallingConventionWrapper` here?
https://github.com/jordanbriere/Source.Python/blob/6d26bca14429b498e5862ccbbfbae57fd0bb7c8a/src/core/modules/memory/memory_function.cpp#L199-L200

> > I couldn't understand the problem. Is this about general documentation? Or is it about the docstring?
> 
> The problem isn't about documentation, but that `CallingConvention` should be self-initializing and shouldn't rely on `CFunction` to be fully functional. The following addresses that: [282bf03](https://github.com/CookStar/Source.Python/compare/add_default_callingconvention...jordanbriere:add_default_callingconvention)
> 
> It also change the retrieval priority of `default_convention` which should be a fallback if no default convention was passed on construction of the instance itself. Basically, this:
> 
> ```python
> class Conv(CallingConvention):
>     default_convention = Convention.THISCALL
> 
> conv = Conv(..., default_convention=Convention.CDECL)
> ```
> 
> Should prioritize `CDECL` because it was explicitly given on instantiation of the instance, as opposed to be set to the class itself so it follows Python resolution order policies.

I completely understand, I think this is the proper implementation.

For the record, in the first implementation, instead of using the existing `ICallingConventionWrapper`, I made a separate class template for `ICallingConvention` and created a class that inherits the default Calling Convention(`x86MsCdecl`, etc.). However, I thought that the binary would become too big for the functionality, so I implemented it as it is now, but I think this post-construction initialization is a necessary feature in any case.

jordanbriere: > I have a little question, why are you checking `ICallingConventionWrapper` here?

Checking `m_eCallingConvention` should be enough, but since it is declared as public, it could end up being externally mutated at some point which would potentially introduce hard to diagnose issues if the convention is being managed by both; `DeleteHook` and `ICallingConventionWrapper`. The `dynamic_cast` really just add extra security to ensure we are indeed working with a built-in convention.

> I understand completely now, I think this is the proper implementation.
> 
> For the record, in the first implementation, instead of using the existing `ICallingConventionWrapper`, I made a separate class template for `ICallingConvention` and created a class that inherits the default Calling Convention(`x86MsCdecl`, etc.). However, I thought that the binary would become too big for the functionality, so I implemented it as it is now, but I think this post-construction initialization is a necessary feature in any case.

Yes, I think this is a great addition in general. Though I don't really like the current implementation as it adds an extra layer and force us to re-pack the arguments twice. I think stackable policies could probably be better. 🤔

CookStar: > Checking `m_eCallingConvention` should be enough, but since it is declared as public, it could end up being externally mutated at some point which would potentially introduce hard to diagnose issues if the convention is being managed by both; `DeleteHook` and `ICallingConventionWrapper`. The `dynamic_cast` really just add extra security to ensure we are indeed working with a built-in convention.

Okay.



> Yes, I think this is a great addition in general. Though I don't really like the current implementation as it adds an extra layer and force us to re-pack the arguments twice. I think stackable policies could probably be better. 
> 
> **EDIT:** This sure sounds like a better approach: [98d9853](https://github.com/jordanbriere/Source.Python/commit/98d985348d1269a745e751eb8475be85f5500039)

Nice! One small error is that in my environment, I get an ambiguous error if I don't have boost::python with `detail` and `make_tuple`.

jordanbriere: > Nice! One small error is that in my environment, I get an ambiguous error if I don't have boost::python with `detail` and `make_tuple`.

Feel free to add them, it is likely an header that includes `wrap_macros.h` while also including [in]directly from a `boost/python/detail/*`'s file that defines into the `detail` scope consequently clashing with our `using nam...`.

CookStar: > Feel free to add them, it is likely an header that includes `wrap_macros.h` while also including [in]directly from a `boost/python/detail/*`'s file that defines into the `detail` scope consequently clashing with our `using nam...`.

Okey. https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/f651f5785bd80af6ee3285b96ec4b808fddbeca5

I think the default convention now works perfectly! Thank you @jordanbriere!

jordanbriere:
Alright, I ran some final tests on this and everything appears to be working as intended. There are just a small improvement I would like to suggest regarding to how overrides are validated. For example, the following:

https://github.com/Source-Python-Dev-Team/Source.Python/blob/f651f5785bd80af6ee3285b96ec4b808fddbeca5/src/core/modules/memory/memory_wrap.h#L121-L125

Is currently validating `get_registers` twice; once through the `!` operator, then another through `CHECK_OVERRIDE`. We should remove the later and validate ourselves instead. For example:

```cpp
		override get_registers = get_override("get_registers");
		if (get_registers.is_none())
		{
			if (m_pDefaultCallingConvention)
				return m_pDefaultCallingConvention->GetRegisters();
			else
				BOOST_RAISE_EXCEPTION(
					PyExc_NotImplementedError,
					"Calling conventions that do not have a default convention "
					"must implement a 'get_registers' method."
				)
		}
```

That way the existence of an override is only tested once. But also, it makes the exception explicit as to which override is missing because a `Method must be implemented by a subclass.` is quite too generic here because `DynamicHooks` is initiating the calls so we don't have any guidance through the Python traceback to tell us which implementation is missing.

Once that is addressed for all the methods, and assuming @Ayuto does not have extra points he would like to be addressed, then I think it is in a good state to be merged.

CookStar:
> There are just a small improvement I would like to suggest regarding to how overrides are validated. For example, the following:
> 
> https://github.com/Source-Python-Dev-Team/Source.Python/blob/f651f5785bd80af6ee3285b96ec4b808fddbeca5/src/core/modules/memory/memory_wrap.h#L121-L125
> 
> Is currently validating `get_registers` twice; once through the `!` operator, then another through `CHECK_OVERRIDE`. We should remove the later and validate ourselves instead. For example:
> 
> ```c++
> 		override get_registers = get_override("get_registers");
> 		if (get_registers.is_none())
> 		{
> 			if (m_pDefaultCallingConvention)
> 				return m_pDefaultCallingConvention->GetRegisters();
> 			else
> 				BOOST_RAISE_EXCEPTION(
> 					PyExc_NotImplementedError,
> 					"Calling conventions that do not have a default convention "
> 					"must implement a 'get_registers' method."
> 				)
> 		}
> ```
> 
> That way the existence of an override is only tested once. But also, it makes the exception explicit as to which override is missing because a `Method must be implemented by a subclass.` is quite too generic here because `DynamicHooks` is initiating the calls so we don't have any guidance through the Python traceback to tell us which implementation is missing.

That's a good idea!

Should I delete this obsolete CHECK_OVERRIDE?

https://github.com/Source-Python-Dev-Team/Source.Python/blob/f651f5785bd80af6ee3285b96ec4b808fddbeca5/src/core/utilities/wrap_macros.h#L208-L213

jordanbriere:
> Should I delete this obsolete CHECK_OVERRIDE?

Hmm. I think it can be removed. It is too generic in most cases, and I think an utility functions (into [sp_utils.h](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/utilities/sp_util.h)) like this:

```cpp
inline override GetOverride(const char *szName)
{
	override oOverride = get_override(szName);
	if (oOverride.is_none())
		BOOST_RAISE_EXCEPTION(
			PyExc_NotImplementedError,
			"Method '%s' must be implemented by a subclass.",
			szName
		)

	return oOverride;
}
```

Or similar, could be more useful than a macro. It could be used when there is not a need to do anything else when an override is missing. For example, there:

https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/engines/engines.h#L125
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/engines/engines.h#L130
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/engines/engines.h#L143
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/events/events.h#L129

So that they are more informational than their current `NoneType not callable`, etc.

CookStar:
> **EDIT:** Scrap that, it's a member method and is protected. Perhaps a macro that retrieve and raise could works better. Anyways, not really relevant to the PR at hand.

Just to clarify, which one do you mean by `it's a member method and is protected`?

jordanbriere:
> 
> 
> > **EDIT:** Scrap that, it's a member method and is protected. Perhaps a macro that retrieve and raise could works better. Anyways, not really relevant to the PR at hand.
> 
> Just to clarify, which one do you mean by `it's a member method and is protected`?

I was talking about `get_override`, just forget about it. :)

Ayuto:
>Once that is addressed for all the methods, and assuming @Ayuto does not have extra points he would like to be addressed, then I think it is in a good state to be merged.

I don't have anything to add. Thank you both of you for your effort!

PR #345, opened 2020-08-08
Fixed a typo in one of the CFogController inputs.

vinci6k:
Took me a while to realize why `AttributeError: Attribute "set_color_lerp_to" not found` kept popping up.


Ayuto:
Thank you!

PR #348, opened 2020-09-01
Add Fastcall calling convention.

CookStar:
This adds a Fastcall calling convention. Ayuto/DynamicHooks#6
Fastcall is used by some CS:GO functions,  but one problem is that there are functions that require caller to clean the stack.

This can be handled by default_convention(#344).


Ayuto:
Thank you for the PR (also over there in the DynamicHooks repo)! Unfortunately, I'm won't have time to review and create a new build before october. If you urgently need the convention, you can also implement it via Python in you plugin and replace it with the built-in convention as soon as we have merged the PR and created a new build.

Is there any reason why you added the cpp-file? That should be part of the compiled DynamicHooks lib.

CookStar:
> If you urgently need the convention, you can also implement it via Python in you plugin and replace it with the built-in convention as soon as we have merged the PR and created a new build.

It's not a sudden need for me, but I'm making that suggestion because I think the expansion of Source.Python's functionality is necessary.

> Is there any reason why you added the cpp-file? That should be part of the compiled DynamicHooks lib.

I included cpp-file for consistency, as cpp-file from other calling conventions has been included as well.

Ayuto:
>It's not a sudden need for me, but I'm making that suggestion because I think the expansion of Source.Python's functionality is necessary.

Yes, definitely! And I'm happy you are doing all the PRs :)

>I included cpp-file for consistency, as cpp-file from other calling conventions has been included as well.

No, they aren't. Did you perhaps look into the DynamicHooks repo instead of the repo of SP?

CookStar:
> No, they aren't. Did you perhaps look into the DynamicHooks repo instead of the repo of SP?

https://github.com/Source-Python-Dev-Team/Source.Python/tree/master/src/thirdparty/DynamicHooks/include/conventions

Ayuto:
Oops, I looked into the previous directory, where all cpp files have been removed. That's also how it should be done with the conventions.

CookStar:
> Oops, I looked into the previous directory, where all cpp files have been removed. That's also how it should be done with the conventions.

Done.

Ayuto:
Thank you!

PR #349, opened 2020-09-17
Add trampoline property to CFunction.

CookStar:
This change allows hooked function to be called dynamically from other function libraries.

### DIFF
--- a/src/core/modules/memory/memory_function.cpp
+++ b/src/core/modules/memory/memory_function.cpp
@@ -321,6 +321,16 @@ object CFunction::Call(tuple args, dict kw)
 	return object();
 }
 
+object CFunction::CallTrampoline(tuple args, dict kw)
+{
+	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
+	if (!pHook)
+		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function was not hooked.")
+
+	return CFunction((unsigned long) pHook->m_pTrampoline, m_eCallingConvention,
+		m_iCallingConvention, m_tArgs, m_eReturnType, m_oConverter).Call(args, kw);
+}
+
 object CFunction::SkipHooks(tuple args, dict kw)
 {
 	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
--- a/src/core/modules/memory/memory_function.cpp
+++ b/src/core/modules/memory/memory_function.cpp
@@ -337,10 +337,10 @@ object CFunction::SkipHooks(tuple args, dict kw)
 unsigned long CFunction::GetTrampolineAddress()
 {
 	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
-	if (pHook)
-		return (unsigned long) pHook->m_pTrampoline;
+	if (!pHook)
+		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function was not hooked.")
 
-	return m_ulAddr;
+	return (unsigned long) pHook->m_pTrampoline;
 }
 
 CHook* HookFunctionHelper(void* addr, ICallingConvention* pConv)
--- a/src/core/modules/memory/memory_function.cpp
+++ b/src/core/modules/memory/memory_function.cpp
@@ -170,13 +170,13 @@ CFunction::CFunction(unsigned long ulAddr, object oCallingConvention, object oAr
 }
 
 CFunction::CFunction(unsigned long ulAddr, Convention_t eCallingConvention,
-	int iCallingConvention, ICallingConvention* pCallingConvention, tuple tArgs,
-	DataType_t eReturnType, object oConverter)
+	int iCallingConvention, tuple tArgs, DataType_t eReturnType, object oConverter)
 	:CPointer(ulAddr)
 {
 	m_eCallingConvention = eCallingConvention;
 	m_iCallingConvention = iCallingConvention;
-	m_pCallingConvention = pCallingConvention;
+	m_pCallingConvention = NULL;
+	m_oCallingConvention = object();
 
 	// We didn't allocate the calling convention, someone else is responsible for it.
 	m_bAllocatedCallingConvention = false;
@@ -216,6 +216,16 @@ bool CFunction::IsHookable()
 bool CFunction::IsHooked()
 {
 	return GetHookManager()->FindHook((void *) m_ulAddr) != NULL;
+}
+
+CFunction* CFunction::GetTrampoline()
+{
+	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
+	if (!pHook)
+		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function was not hooked.")
+
+	return new CFunction((unsigned long) pHook->m_pTrampoline, m_eCallingConvention,
+		m_iCallingConvention, m_tArgs, m_eReturnType, m_oConverter);
 }
 
 template<class ReturnType, class Function>
@@ -312,35 +322,14 @@ object CFunction::Call(tuple args, dict kw)
 	return object();
 }
 
-object CFunction::CallTrampoline(tuple args, dict kw)
-{
-	if (!IsCallable())
-		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function is not callable.")
-
-	Validate();
-	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
-	if (!pHook)
-		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function was not hooked.")
-
-	return CFunction((unsigned long) pHook->m_pTrampoline, m_eCallingConvention,
-		m_iCallingConvention, m_pCallingConvention, m_tArgs, m_eReturnType, m_oConverter).Call(args, kw);
-}
-
 object CFunction::SkipHooks(tuple args, dict kw)
 {
-	if (IsHooked())
-		return CallTrampoline(args, kw);
+	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
+	if (pHook)
+		return CFunction((unsigned long) pHook->m_pTrampoline, m_eCallingConvention,
+			m_iCallingConvention, m_tArgs, m_eReturnType, m_oConverter).Call(args, kw);
 
 	return Call(args, kw);
-}
-
-unsigned long CFunction::GetTrampolineAddress()
-{
-	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
-	if (!pHook)
-		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Function was not hooked.")
-
-	return (unsigned long) pHook->m_pTrampoline;
 }
 
 CHook* HookFunctionHelper(void* addr, ICallingConvention* pConv)
--- a/src/core/modules/memory/memory_function.cpp
+++ b/src/core/modules/memory/memory_function.cpp
@@ -176,7 +176,6 @@ CFunction::CFunction(unsigned long ulAddr, Convention_t eCallingConvention,
 	m_eCallingConvention = eCallingConvention;
 	m_iCallingConvention = iCallingConvention;
 	m_pCallingConvention = NULL;
-	m_oCallingConvention = object();
 
 	// We didn't allocate the calling convention, someone else is responsible for it.
 	m_bAllocatedCallingConvention = false;
--- a/src/core/modules/memory/memory_function.cpp
+++ b/src/core/modules/memory/memory_function.cpp
@@ -334,6 +334,15 @@ object CFunction::SkipHooks(tuple args, dict kw)
 	return Call(args, kw);
 }
 
+unsigned long CFunction::GetTrampolineAddress()
+{
+	CHook* pHook = GetHookManager()->FindHook((void *) m_ulAddr);
+	if (pHook)
+		return (unsigned long) pHook->m_pTrampoline;
+
+	return m_ulAddr;
+}
+
 CHook* HookFunctionHelper(void* addr, ICallingConvention* pConv)
 {
 	CHook* result;
--- a/src/core/modules/memory/memory_function.h
+++ b/src/core/modules/memory/memory_function.h
@@ -66,20 +66,22 @@ public:
 	bool IsHookable();
 
 	bool IsHooked();
-    
+
 	object Call(boost::python::tuple args, dict kw);
 	object CallTrampoline(boost::python::tuple args, dict kw);
 	object SkipHooks(boost::python::tuple args, dict kw);
-	
+
+	unsigned long GetTrampolineAddress();
+
 	void AddHook(HookType_t eType, PyObject* pCallable);
 	void RemoveHook(HookType_t eType, PyObject* pCallable);
-    
+
 	void AddPreHook(PyObject* pCallable)
 	{ return AddHook(HOOKTYPE_PRE, pCallable); }
 
 	void AddPostHook(PyObject* pCallable)
 	{ return AddHook(HOOKTYPE_POST, pCallable); }
-    
+
 	void RemovePreHook(PyObject* pCallable)
 	{ RemoveHook(HOOKTYPE_PRE, pCallable); }
 
@@ -87,7 +89,7 @@ public:
 	{ RemoveHook(HOOKTYPE_POST, pCallable);	}
 
 	void DeleteHook();
-    
+
 public:
 	boost::python::tuple	m_tArgs;
 	object					m_oConverter;
--- a/src/core/modules/memory/memory_function.h
+++ b/src/core/modules/memory/memory_function.h
@@ -69,6 +69,7 @@ public:
 	CFunction* GetTrampoline();
 
 	object Call(boost::python::tuple args, dict kw);
+	object CallTrampoline(boost::python::tuple args, dict kw);
 	object SkipHooks(boost::python::tuple args, dict kw);
 
 	void AddHook(HookType_t eType, PyObject* pCallable);
--- a/src/core/modules/memory/memory_function.h
+++ b/src/core/modules/memory/memory_function.h
@@ -57,8 +57,7 @@ class CFunction: public CPointer, private boost::noncopyable
 public:
 	CFunction(unsigned long ulAddr, object oCallingConvention, object oArgs, object oReturnType);
 	CFunction(unsigned long ulAddr, Convention_t eCallingConvention, int iCallingConvention,
-		ICallingConvention* pCallingConvention, boost::python::tuple tArgs,
-		DataType_t eReturnType, object oConverter);
+		boost::python::tuple tArgs, DataType_t eReturnType, object oConverter);
 
 	~CFunction();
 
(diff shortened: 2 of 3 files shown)


jordanbriere:
> 
> 
> This change allows hooked function to be called dynamically from other function libraries.

What other libraries? Can't they use `Function.call_trampoline` to call it?

CookStar:
> What other libraries?

The most effective one is ctypes.

Function call using dyncall have a definite overhead and also have the disadvantage not being able to call a function for a certain calling convention. This commit fixes these problems.

jordanbriere:
> 
> 
> > What other libraries?
> 
> The most effective one is ctypes.
> 
> Function call using dyncall have a definite overhead and also have the disadvantage not being able to call a function with a certain calling convention. This commit allows you to fix these problems.

Hmm. Fair enough. Though, what tests did you run to compare both? And what was your performance metrics? I never dynamically called function pointers with ctypes, and I'm really curious about its performance.

> "Return the trampoline address if the function is hooked, otherwise return the function address."

I think it would be better to raise a `ValueError` if the function isn't hooked. Both for consistency, and explicitness. I mean, if someone retrieve the `trampoline_address` property, they expect to get the address of the trampoline, not the address of the function.

CookStar:
> Hmm. Fair enough. Though, what tests did you run to compare both? And what was your performance metrics? I never dynamically called function pointers with ctypes, and I'm really curious about its performance.

This is what it looks like in the testing of an extreme case.
```python
import ctypes
import timeit

from memory import Convention
from memory import DataType
from memory import Pointer

engine = find_binary("engine", srv_check=False)
NET_BufferToBufferCompress = engine[b"\x55\x89\xE5\x83\xEC\x48\x89\x5D\xF4\x8B\x5D\x14\x89\x75\xF8\x8B\x75\x10"].make_function(
    Convention.CDECL, (
        DataType.POINTER,
        DataType.POINTER,
        DataType.POINTER,
        DataType.UINT,
    ),
    DataType.BOOL,
)

functype = ctypes.CFUNCTYPE(
    ctypes.c_bool,
    ctypes.c_void_p,
    ctypes.c_void_p,
    ctypes.c_void_p,
    ctypes.c_uint32
)
c_NET_BufferToBufferCompress = functype(NET_BufferToBufferCompress.address)

loop = 100000
data = b"\x00"

stmt_sp = """

input_length = len(data)

input = (ctypes.c_char*input_length).from_buffer_copy(data)
input_address = ctypes.addressof(input)
input_pointer = Pointer(input_address)

output = bytearray(input_length)
output_address = ctypes.addressof(ctypes.c_char.from_buffer(output))
output_pointer = Pointer(output_address)

output_length = ctypes.c_uint32(input_length)
output_length_address = ctypes.addressof(output_length)
output_length_pointer = Pointer(output_length_address)

result = NET_BufferToBufferCompress(output_pointer, output_length_pointer, input_pointer, input_length)

"""
result = timeit.timeit(stmt_sp, globals=globals(), number=loop)
print("sp", result)


stmt_ctypes = """

input_length = len(data)

output_data = bytearray(input_length)
output = ctypes.c_char.from_buffer(output_data)

output_length = ctypes.c_uint32(input_length)

result = c_NET_BufferToBufferCompress(ctypes.byref(output), ctypes.byref(output_length), data, input_length)

"""
result = timeit.timeit(stmt_ctypes, globals=globals(), number=loop)
print("ctypes", result)
```

Output:
```
sp 4.656123496592045
ctypes 0.19711758010089397
```

In the real world, the function `call time` can be reduced by about half. (There is an advantage of passing the bytes directly to the function, though.)
This really comes in handy with functions that call multiple times per tick, especially when you can reuse the pointer.

> I think it would be better to raise a `ValueError` if the function isn't hooked. Both for consistency, and explicitness. I mean, if someone retrieve the `trampoline_address` property, they expect to get the address of the trampoline, not the address of the function.

I implemented it in the same way as SkipHooks, but that's probably better.

Ayuto:
We could also expose a ``trampoline`` property that returns ``CFunction`` instance. Then we can simply call it using SP or do ``func.trampoline.address`` to handle the calls with ctypes. Though, my initial idea was to not directly expose the trampoline address, so people don't try to hook the trampoline. If we expose the ``trampoline`` property, we should deprecate ``call_trampoline``.

CookStar:
> We could also expose a `trampoline` property that returns `CFunction` instance. Then we can simply call it using SP or do `func.trampoline.address` to handle the calls with ctypes. Though, my initial idea was to not directly expose the trampoline address, so people don't try to hook the trampoline. If we expose the `trampoline` property, we should deprecate `call_trampoline`.

If we could deprecate `call_trampoline`, then implementing `trampoline property` is the best way to go! https://github.com/Source-Python-Dev-Team/Source.Python/pull/349/commits/ad6d71854b973f327cd08ded3aae0e97c830bd46

jordanbriere:
I like the idea of the `trampoline` property. If the trampoline address is not subject to change, we could also declare it as cached so that we don't initialize a new one every time. As for deprecating `call_trampoline` I have a mixed feeling about that. There are a few snippets on the forums that use it, and this also means updating a few classes internally (`MemberFunction` and `EntityMemFunc` comes to mind). I think I would rather vote to keep it along the property for backward compatibility. The same way we have `Entity.get_input` and `Entity.call_input` for convenience, etc.

CookStar:
> As for deprecating `call_trampoline` I have a mixed feeling about that. There are a few snippets on the forums that use it, and this also means updating a few classes internally (`MemberFunction` and `EntityMemFunc` comes to mind). I think I would rather vote to keep it along the property for backward compatibility. The same way we have `Entity.get_input` and `Entity.call_input` for convenience, etc.

Reverted the deletion of call_trampoline. https://github.com/Source-Python-Dev-Team/Source.Python/pull/349/commits/8c28170e8eade372a0bb6bef190bb4f0ee56926e

jordanbriere:
Quick question; any specific reason why `pCallingConvention` was removed from the constructor?

CookStar:
https://github.com/Source-Python-Dev-Team/Source.Python/blob/8c28170e8eade372a0bb6bef190bb4f0ee56926e/src/core/modules/memory/memory_function.cpp#L210-L213

The trampoline function will not be hooked so there is no need for `pCallingConvention`.

jordanbriere:
> 
> 
> https://github.com/Source-Python-Dev-Team/Source.Python/blob/8c28170e8eade372a0bb6bef190bb4f0ee56926e/src/core/modules/memory/memory_function.cpp#L210-L213
> 
> The trampoline function will not be hooked so there is no need for `pCallingConvention`.

Yes, I get that. I'm just not sure why removing it from the constructor is preferred over passing `NULL` to it.

CookStar:
Since this CFunction only generates unhookable functions, I decided that `pCallingConvention` is not necessary. And it's just a small optimization.

jordanbriere:
Quick update regarding this:

> Function call using dyncall have a definite overhead

I ran some tests just now and the bottleneck wasn't `DynCall` but actually the way we were internally extracting pointers. Basically, the nested-inlining of try/catch along with the order we were looking for pointers into `ExtractPointer` was causing Python's exception handler to be initialized and cleared multiple times which doesn't make much sense in the context of an extraction that do not need to invoke Python whatsoever. This was caused by the fact we were first trying to dynamically resolve the pointer of the object (through `_ptr()`) while the priority should be to expect a `CPointer` first. Anyways, should be much faster into: https://github.com/Source-Python-Dev-Team/Source.Python/commit/5cbb96896df06991a27871383184dc9aee0ca68c

Code I used to test:

```python
from time import time

from memory import *
from entities.entity import *
from mathlib import *

world = Entity(0)
this = world.pointer

teleport = Function(
    world.teleport.address,
    world.teleport.convention,
    world.teleport.arguments,
    world.teleport.return_type
)

vector = Vector()
ptr = get_object_pointer(Vector())

t = time()
for i in range(1000000):
    teleport(this, ptr, ptr, ptr)
print('Pointers:', time() - t)

t = time()
for i in range(1000000):
    teleport(world, vector, vector, vector)
print('Objects:', time() - t)
```

Results:

```python
Before:
    Pointers: 117.46061897277832
    Objects: 4.552911758422852

After:
    Pointers: 1.3749786472320557
    Objects: 5.238690376281738
```

Resolving objects is slightly slower, because we now first look for pointers but calling with pointers directly is about ~85 times faster and should have a major impact on the metrics you got above.

CookStar:
> I ran some tests just now and the bottleneck wasn't `DynCall` but actually the way we were internally extracting pointers. Basically, the nested-inlining of try/catch along with the order we were looking for pointers into `ExtractPointer` was causing Python's exception handler to be initialized and cleared multiple times which doesn't make much sense in the context of an extraction that do not need to invoke Python whatsoever. This was caused by the fact we were first trying to dynamically resolve the pointer of the object (through `_ptr()`) while the priority should be to expect a `CPointer` first. Anyways, should be much faster into: [5cbb968](https://github.com/Source-Python-Dev-Team/Source.Python/commit/5cbb96896df06991a27871383184dc9aee0ca68c)

You are absolutely right!
In my environment, there was roughly a 35x performance increase.

However, ctypes is still fast. (Roughly twice as fast as I wrote before.)

```
Before(Linux):
    Pointers: 50.75538492202759
    Objects: 4.031508207321167
    Ctypes: 0.735945463180542

After(Linux):
    Pointers: 1.4371249675750732
    Objects: 4.843799829483032
    Ctypes: 0.7265281677246094

Windows(Only ctypes):
    Ctypes: 0.8089907169342041
```

Previous test(Linux).
https://github.com/Source-Python-Dev-Team/Source.Python/pull/349#issuecomment-697311618
```
sp 0.6994638182222843
ctypes 0.20021243020892143
```

The test code was changed because there were a lot of `Bad SetLocalAngles` errors on Linux.
Test Code(Linux):
```python
from time import time

from memory import *
from entities.entity import *
from mathlib import *

# https://github.com/CookStar/SmallLib/tree/master/addons/source-python/packages/custom/memorytools
from memorytools.ctypes import get_ctype_function

world = Entity(0)
this = world.pointer

teleport = Function(
    world.teleport.address,
    world.teleport.convention,
    world.teleport.arguments,
    world.teleport.return_type
)

vector = Vector()
ptr = get_object_pointer(Vector())
vector2 = Vector()
ptr2 = get_object_pointer(Vector())
vector3 = Vector()
ptr3 = get_object_pointer(Vector())

t = time()
for i in range(1000000):
    teleport(this, ptr, ptr2, ptr3)
print('Pointers:', time() - t)

vector = Vector()
ptr = get_object_pointer(Vector())
vector2 = Vector()
ptr2 = get_object_pointer(Vector())
vector3 = Vector()
ptr3 = get_object_pointer(Vector())

t = time()
for i in range(1000000):
    teleport(world, vector, vector2, vector3)
print('Objects:', time() - t)

vector = Vector()
ptr = get_object_pointer(Vector())
vector2 = Vector()
ptr2 = get_object_pointer(Vector())
vector3 = Vector()
ptr3 = get_object_pointer(Vector())

c_teleport = get_ctype_function(teleport, auto_dealloc=False).ctype
c_world_p = ctypes.c_void_p(this.address)
c_vector_p = ctypes.c_void_p(ptr.address)
c_vector_p2 = ctypes.c_void_p(ptr2.address)
c_vector_p3 = ctypes.c_void_p(ptr3.address)

t = time()
for i in range(1000000):
    c_teleport(c_world_p, c_vector_p, c_vector_p, c_vector_p)
print('Ctypes:', time() - t)
```

On Windows, there are problems with arguments, so I added Bool.
Also, I can't compile on Windows right now, so it's only ctypes.

```python
from time import time

from memory import *
from entities.entity import *
from mathlib import *

# https://github.com/CookStar/SmallLib/tree/master/addons/source-python/packages/custom/memorytools
from memory import DataType
from memorytools.ctypes import get_ctype_calling_convention
from memorytools.ctypes import get_ctype_argtypes
from memorytools.ctypes import get_ctype_from_data_type

world = Entity(0)
this = world.pointer

teleport = Function(
    world.teleport.address,
    world.teleport.convention,
    world.teleport.arguments,
    world.teleport.return_type
)

vector = Vector()
ptr = get_object_pointer(Vector())
teleport(world, vector, vector, vector)

convention = get_ctype_calling_convention(teleport.convention)
arguments = list(get_ctype_argtypes(teleport.arguments))
arguments.append(get_ctype_from_data_type(DataType.BOOL))
return_type = get_ctype_from_data_type(teleport.return_type)

c_teleport = convention(teleport.address, arguments, return_type, auto_dealloc=False).ctype

c_world_p = ctypes.c_void_p(this.address)
c_vector_p = ctypes.c_void_p(ptr.address)

t = time()
for i in range(1000000):
    c_teleport(c_world_p, c_vector_p, c_vector_p, c_vector_p, True)
print('Ctypes:', time() - t)
```

Whether this difference is necessary or not is up to the user, but ctypes can also support calling conventions that are not supported by `DynCall`.

```python
import ctypes
import timeit

from memory import Pointer

from memorytools.conventions import FASTCALL_CALLER
from memorytools.ctypes import get_ctype_function

engine = find_binary("engine", srv_check=False)
NET_BufferToBufferCompress = engine[b"\x55\x8B\xEC\x83\xE4\xF8\x81\xEC\xE4\x00\x00\x00\x53\x56\x57"].make_function(
    FASTCALL_CALLER, (
        DataType.POINTER,
        DataType.POINTER,
        DataType.POINTER,
        DataType.UINT,
    ),
    DataType.BOOL,
)

c_NET_BufferToBufferCompress = get_ctype_function(NET_BufferToBufferCompress, auto_dealloc=False).ctype

def compress_file(path, data):
    input_length = len(data)

    output_data = bytearray(input_length)
    output = ctypes.c_char.from_buffer(output_data)

    output_length = ctypes.c_uint32(input_length)

    if c_NET_BufferToBufferCompress(ctypes.byref(output), ctypes.byref(output_length), data, input_length):
        ztmp_path = path.with_suffix(path.suffix+".ztmp")
        with open(ztmp_path, "wb") as file:
            file.write(output_data[0:output_length.value])
        ztmp_path.chmod(0o755)
```

This is something you can't do with the normal Source.Python functionality.

jordanbriere:
> Previous test(Linux).
> https://github.com/Source-Python-Dev-Team/Source.Python/pull/349#issuecomment-697311618
> ```
> sp 0.6994638182222843
> ctypes 0.20021243020892143
> ```

Much better! ^^

> Whether this difference is necessary or not is up to the user

There are ways `CFunction::Call` could be optimized directly as well. For example, the following two switches:

https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_function.cpp#L267
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_function.cpp#L297

Could easily be removed. I mean, we know the arguments on construction, and all the cases does the same thing; extract and forward based on type so mapping templated dispatchers that we simply loop through and call sounds easy enough and would remove the need to test the types every call for every arguments. 🤔 

> https://github.com/CookStar/SmallLib/tree/master/addons/source-python/packages/custom/memorytools

Nice library! I will definitely play around with it when I get the chance and see what it has to offer.

CookStar:
> There are ways `CFunction::Call` could be optimized directly as well. For example, the following two switches:
> 
> https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_function.cpp#L267
> https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_function.cpp#L297
> 
> Could easily be removed. I mean, we know the arguments on construction, and all the cases does the same thing; extract and forward based on type so mapping templated dispatchers that we simply loop through and call sounds easy enough and would remove the need to test the types every call for every arguments.

It certainly looks like it could be further improved.

> Nice library! I will definitely play around with it when I get the chance and see what it has to offer.

This is a library that provides specific custom conventions and execution by ctypes.
When it's ready, I'll post it on the forum.

jordanbriere:
> This is a library that provides specific custom conventions and execution by ctypes.
> When it's ready, I'll post it on the forum.

Cool, I will have a look then. Slight note:

https://github.com/CookStar/SmallLib/blob/master/addons/source-python/packages/custom/memorytools/ctypes.py#L196-L230

It is usually much better to use a dictionary, because long functions with ton of elif's can prove to be very inefficient. For example, something like would be much better due to hashing:

```python
_data_type_to_ctype = {
    DataType.VOID: None,
    DataType.BOOL: ctypes.c_bool,
    DataType.CHAR: ctypes.c_char,
    DataType.UCHAR: ctypes.c_ubyte,
    DataType.SHORT: ctypes.c_short,
    DataType.USHORT: ctypes.c_ushort,
    DataType.INT: ctypes.c_int,
    DataType.UINT: ctypes.c_uint,
    DataType.LONG: ctypes.c_long,
    DataType.ULONG: ctypes.c_ulong,
    DataType.LONG_LONG: ctypes.c_longlong,
    DataType.ULONG_LONG: ctypes.c_ulonglong,
    DataType.FLOAT: ctypes.c_float,
    DataType.DOUBLE: ctypes.c_double,
    DataType.POINTER: ctypes.c_void_p,
    DataType.STRING: ctypes.c_char_p
}

def get_ctype_from_data_type(data_type):
    try:
        return _data_type_to_ctype[data_type]
    except KeyError:
        raise ValueError("Given data_type is not supported.")
```

You could be surprised to time the differences!

CookStar:
> It is usually much better to use a dictionary, because long functions with ton of elif's can prove to be very inefficient. For example, something like this:
> 
> ```python
> _data_type_to_ctype = {
>     DataType.VOID: None,
>     DataType.BOOL: ctypes.c_bool,
>     DataType.CHAR: ctypes.c_char,
>     DataType.UCHAR: ctypes.c_ubyte,
>     DataType.SHORT: ctypes.c_short,
>     DataType.USHORT: ctypes.c_ushort,
>     DataType.INT: ctypes.c_int,
>     DataType.UINT: ctypes.c_uint,
>     DataType.LONG: ctypes.c_long,
>     DataType.ULONG: ctypes.c_ulong,
>     DataType.LONG_LONG: ctypes.c_longlong,
>     DataType.ULONG_LONG: ctypes.c_ulonglong,
>     DataType.FLOAT: ctypes.c_float,
>     DataType.DOUBLE: ctypes.c_double,
>     DataType.POINTER: ctypes.c_void_p,
>     DataType.STRING: ctypes.c_char_p
> }
> 
> def get_ctype_from_data_type(data_type):
>     try:
>         return _data_type_to_ctype[data_type]
>     except KeyError:
>         raise ValueError("Given data_type is not supported.")
> ```
> 
> Would be much better due to hashing. You could be surprised to time the differences!

```python
import time

from memory import Convention
from memory import DataType

from memorytools.ctypes import get_ctype_function

engine = find_binary("engine", srv_check=False)
NET_BufferToBufferCompress = engine[b"\x55\x89\xE5\x83\xEC\x48\x89\x5D\xF4\x8B\x5D\x14\x89\x75\xF8\x8B\x75\x10"].make_function(
    Convention.CDECL, (
        DataType.POINTER,
        DataType.POINTER,
        DataType.POINTER,
        DataType.UINT,
    ),
    DataType.BOOL,
)

s = time.perf_counter()
for i in range(1000000):
    c_func = get_ctype_function(NET_BufferToBufferCompress)
e = time.perf_counter()
print(e-s)
```
Output:
```
Before: 8.506793942302465
After: 5.99107801169157
```
Indeed, yes!
Although, it doesn't affect the execution of the ctypes function itself, since it's just for generating the ctypes function.
Thanks for your teachings!

PR #351, opened 2020-09-22
Added networked variants of the entity listeners.

jordanbriere:
This branch adds networked variants of the entity listeners. The rationale behind this is that, 9 (if not 10) times out of 10, those listeners are used like so:

```python
@OnEntitySomething
def on_entity_something(base_entity):
    try:
        index = base_entity.index
    except ValueError:
        return

    ...


@OnEntitySomethingElse
def on_entity_something_else(base_entity):
    if not base_entity.is_networked():
        return

    ...
```

Which is redundant. That branch adds the following variants that only get called for networked entities:

```python
@OnNetworkedEntityPreSpawned
def on_networked_entity_pre_spawned(entity):
    ...


@OnNetworkedEntitySpawned
def on_networked_entity_spawned(entity):
    ...


@OnNetworkedEntityCreated
def on_networked_entity_created(entity):
    ...


@OnNetworkedEntityDeleted
def on_networked_entity_deleted(entity):
    ...
```

Allowing for shorter codes, less calls, as well as removing the needs for every callbacks to validate themselves that they are working with networked entities, etc.


Ayuto:
Instead of just passing the index we could also pass the BaseEntity instance or maybe the Entity instance?

jordanbriere:
> 
> 
> Instead of just passing the index we could also pass the BaseEntity instance or maybe the Entity instance?

I passed the index mainly because this is what we use internally, and that this is what is used first by the majority of codes on our forums (through custom entity class, or containers, etc.). However, now that you mention it, I think passing an `Entity` is what make the much sense since they are basically our "networked variant" of a `BaseEntity` and allows for direct class name filtering within a callback. Retrieving the index from it should be fast, since it is cached anyways. Will push that change shortly.

PR #355, opened 2020-10-18
Update CS:GO assists offset for Windows/Linux

ericentin:
Hello, thanks for an amazing project.

After seeing some weirdness with getting/setting player assist values in CS:GO, I did some investigation and it appears that the offset for assists has changed. I obtained these from running the snippet of commented code above the assists offset data on a CS:GO server on Windows and Linux:

Windows: `Offset of Player.assists is: 3204`
Linux: `Offset of Player.assists is: 3228`

This PR sets those values. Apologies if I missed something or have made a mistake. Thanks again!


jordanbriere:
Thank you! :)

PR #360, opened 2020-10-23
Fix remove entity listener

roflmuffin:
When browsing the source code I noticed this typo which was binding "AddListenerEntity" to both add and remove entity listener on the python side.

This only gets called on SP unload so changing this probably doesn't actually do much in the end.


Ayuto:
Good catch, thank you! Indeed, it is currently only called at unloading SP.

PR #361, opened 2020-10-24
Added default argument to Send/Deny/RequestFile functions.

CookStar:
This eliminates unnecessary condition checks.

Before:
```python
# Source.Python Imports
#   Core
from core import SOURCE_ENGINE
#   Players

player = Player(1)
if SOURCE_ENGINE in ("blade", "csgo", "l4d2"):
    player.client.net_channel.send_file(file_name, transfer_id, False)
else:
    player.client.net_channel.send_file(file_name, transfer_id)
```
After:
```python
# Source.Python Imports
#   Players

player = Player(1)
player.client.net_channel.send_file(file_name, transfer_id)
```

### DIFF
--- a/src/core/modules/net_channel/net_channel_wrap.cpp
+++ b/src/core/modules/net_channel/net_channel_wrap.cpp
@@ -91,17 +91,47 @@ void export_net_channel(scope _net_channel)
 	);
 
 	NetChannel.def(
+		"set_data_rate",
+		&INetChannel::SetDataRate
+	);
+
+#if defined(ENGINE_BLADE) || defined(ENGINE_CSGO) || defined(ENGINE_LEFT4DEAD2)
+	NetChannel.def(
 		"send_file",
-		&INetChannel::SendFile
+		&INetChannel::SendFile,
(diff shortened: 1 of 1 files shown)


CookStar:
Apparently, INetChannel::SetDataRate got mixed up too.
If it shouldn't be added, I'll delete it.

jordanbriere:
Looking good to me. It's great to have prototype uniformity between games/engines.

> Apparently, INetChannel::SetDataRate got mixed up too.
> If it shouldn't be added, I'll delete it.

Nothing wrong with that. It's actually a step towards the following todo. 😃 

https://github.com/Source-Python-Dev-Team/Source.Python/blob/ea1fecf4c76acdf59688d961651bebb2d50efe2b/src/core/modules/net_channel/net_channel_wrap.cpp#L142

Question is; was it confirmed to be the same prototype on all engines?

CookStar:
> Question is; was it confirmed to be the same prototype/available on all engines?

Which one do you mean, exactly?

There is no default argument in the blade and csgo headers, but there is one in l4d2.
There is no is_replay_demo in the other engines.
And INetChannel::SetDataRate is available in all engines.

jordanbriere (approved):

PR #367, opened 2020-12-04
Updated CS:GO offsets.

CookStar:


CookStar:
If someone has already found it, there's no need to find it again.

https://gist.github.com/CookStar/b890e1cb3d1c68991b45b0c50a28c80e

jordanbriere:
> 
> 
> If someone has already found it, there's no need to find it again, and thanks to AlliedModders/SourceMod.
> 
> https://gist.github.com/CookStar/b890e1cb3d1c68991b45b0c50a28c80e

Nice! That's actually an interesting way to go about it. Perhaps we could do the same and retrieve them that way dynamically. The chances new members are inserted in-between sounds less likely than being declared before them.

CookStar:
> Nice! That's actually an interesting way to go about it. Perhaps we could do the same and retrieve them that way dynamically. The chances new members are inserted in-between sounds less likely than being declared before them.

If so, it would be better to make the offset-related functions more available on the Python side.
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/entities/entities_props.cpp#L115
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/entities/entities_entity.cpp#L254

jordanbriere:
> If so, it would be better to make the offset-related functions more available on the Python side.
> https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/entities/entities_props.cpp#L115
> https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/entities/entities_entity.cpp#L254

Would definitely be good to have more flexibility for those but I don't think we need them here. This is how I would implement it: https://github.com/Source-Python-Dev-Team/Source.Python/compare/instance_property (only tested on CS:S).

It would also give access to their offset directly from an `Entity` instance:

```python
for name, prop in entity.properties.items():
    print(f'Offset of {name} is {prop.offset}')
```

CookStar:
> Would definitely be good to have more flexibility for those but I don't think we need them here. This is how I would implement it: https://github.com/Source-Python-Dev-Team/Source.Python/compare/instance_property (only tested on CS:S).
> 
> It would also give access to their offset directly from an `Entity` instance:
> 
> ```python
> for name, prop in entity.properties.items():
>     print(f'Offset of {name} is {prop.offset}')
> ```

It works perfectly.

PR #370, opened 2020-12-07
Added instance_property support to entity managers.

jordanbriere:
#367

### DIFF
--- a/addons/source-python/data/source-python/entities/csgo/CBasePlayer.ini
+++ b/addons/source-python/data/source-python/entities/csgo/CBasePlayer.ini
@@ -58,7 +58,7 @@ srv_check = False
         offset_windows = 372
 
 
-[instance_attribute]
+[instance_property]
 
 
     # from memory import alloc
@@ -74,8 +74,8 @@ srv_check = False
             # break
     # print('Offset of Player.assists is:', offset)
     [[assists]]
-        offset_windows = 4024
-        offset_linux = 4048
+        base = m_iFrags
+        offset = 4
         type = INT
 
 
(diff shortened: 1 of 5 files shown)


addons/source-python/packages/source-python/entities/classes.py
```diff
@@ -332,6 +333,26 @@ def _get_server_class(self, class_name, datamap):
...
+                )(
+                    Key.as_attribute_type(self, data['type']),
+                    instance.properties[
+                        data.get('base_' + PLATFORM, data.get('base'))
+                    ].offset + Key.as_int(
+                        self, data.get(
+                            'offset_' + PLATFORM, data.get('offset', 0)
+                        )
+                    ),
+                    data.get('doc', None)
```
Ayuto, line 352: Less cryptic calls would be great :D Other than that it looks good to me.

addons/source-python/data/source-python/entities/csgo/CBasePlayer.ini
```diff
@@ -58,7 +58,7 @@ srv_check = False
         offset_windows = 372
 
 
-[instance_attribute]
+[instance_property]
```
Ayuto, line 61: Maybe we should call it ``based_attribute`` or ``based_property`` as you can specify whether it is an ``instance_attribute`` or ``pointer_attribute`` using the ``method`` key.

I think I would prefer ``based_attribute``.

jordanbriere: > 
> 
> Maybe we should call it `based_attribute` or `based_property` as you can specify whether it is an `instance_attribute` or `pointer_attribute` using the `method` key.
> 
> I think I would prefer `based_attribute`.

Alright, I've updated the code.

Ayuto (approved):

PR #371, opened 2020-12-08
Added CSendProxyRecipients function definitions to undefined_symbols.h.

CookStar:
Fixed this error.
```
[Source.Python] Loading...
=========================================================================
[SP-LOADER] Could not load library: bin/core.so
[SP-LOADER] Error Reported: /home/steam/csgo_ds/csgo/addons/source-python/bin/core.so: undefined symbol: _ZN20CSendProxyRecipients12SetRecipientEi
=========================================================================
Failed to load plugin "addons/source-python"
```

Added Parts:
https://github.com/Source-Python-Dev-Team/Source.Python/blob/2669ced199a85559ca1747aeccd62dd342951678/src/patches/csgo/public/dt_send.h#L144-L158


jordanbriere:
> Fixed this error.

Thanks! However, no need to patch the whole file for missing symbols. Just add them to [undefined_symbols.h](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/patches/csgo/undefined_symbols.h). :)

CookStar:
> > Fixed this error.
> 
> Thanks! However, no need to patch the whole file for missing symbols. Just add them to [undefined_symbols.h](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/patches/csgo/undefined_symbols.h). :)

Since all functions are inline functions, this does not work.

Ayuto:
Oh, I should have tested that. Thanks!

But you don't need to make them inline functions. Then adding them to the undefined symbols should work fine :)

CookStar:
> But you don't need to make them inline functions. Then adding them to the undefined symbols should work fine :)

Okay.

Issue #373, opened 2020-12-14
CS:S - Player.drop_weapon crash


Frag1337:
Hello,

when using following snippet while being on a team/alive on the server:

```
from filters.players import PlayerIter

for player in PlayerIter():
    print('Trying to remove weapons from {}..'.format(player.name))

    for weapon in player.weapons():
        print('Found {} from {}. Lets drop it.'.format(weapon.classname, player.name))

        player.drop_weapon(weapon)

        print('Aaand remove it..')

        weapon.remove()
```

I get following crash:

```
sp plugin load test
[SP] Loading plugin 'test'...
Trying to remove weapons from Frag..
Found weapon_knife from Frag. Lets drop it.
Segmentation fault (core dumped)
Add "-debug" to the ./srcds_run command line to generate a debug.log to help with solving this problem
Mon Dec 14 19:40:46 UTC 2020: Server restart in 10 seconds
```

```
sp info

IMPORTANT: Please copy the full output.
--------------------------------------------------------
Checksum      : 17af3a55546e8fed9c8e9b322e817b97
Date          : 2020-12-14 19:43:07.359723
OS            : Linux-5.4.0-58-generic-x86_64-with-debian-bullseye-sid
Game          : css
SP version    : 702
Github commit : 99ed4dca7b439f5fc4ade054355767b23af497df
Server plugins:
   00: Source.Python, (C) 2012-2020, Source.Python Team.
SP plugins:
--------------------------------------------------------
```

The server also crashes on 

`weapon.remove()`

by uncommenting `player.drop_weapon(weapon)`

```
sp plugin load test
[SP] Loading plugin 'test'...
Trying to remove weapons from Frag..
Found weapon_knife from Frag. Lets drop it.
Aaand remove it..
Segmentation fault (core dumped)
Add "-debug" to the ./srcds_run command line to generate a debug.log to help with solving this problem
Mon Dec 14 19:45:04 UTC 2020: Server restart in 10 seconds
```

Not sure if its helpful, but I've used this fix to make Source.Python work:

https://forums.sourcepython.com/viewtopic.php?p=13976#p13976 Method 2

CookStar:
What happens when you install libffi6 using [this method](https://forums.sourcepython.com/viewtopic.php?p=13976#p14827)?
Of course, don't forget to remove the symbolic link to libffi7.

Frag1337:
I 

> 
> 
[...]

Issue #372, opened 2020-12-14
CS:S - Teleport doesnt apply Vector


Frag1337:
Hello,

when using following snippet while being on a team/alive on the server:

```
from filters.players import PlayerIter
from mathlib import Vector

TELEPORT_ORIGIN = Vector(2500, -2500, -1000)
TELEPORT_ANGLE = Vector(90, 0, 0)
TELEPORT_VELOCITY = Vector(0, 0, 2500)

for player in PlayerIter():
    player.teleport(TELEPORT_ORIGIN, TELEPORT_ANGLE, TELEPORT_VELOCITY)
```

I always get teleported to the coordinates 0 0 0. The angle and velocity vector also doesnt do anything then 0 0 0. The snippet works as intended on windows srcds. (Teleports me to the right coordinate and applies angle + velocity)

```
sp info

IMPORTANT: Please copy the full output.
--------------------------------------------------------
Checksum      : 33d8aa51f81ad4cd2e156c8fc993b835
Date          : 2020-12-14 19:20:44.844946
OS            : Linux-5.4.0-58-generic-x86_64-with-debian-bullseye-sid
Game          : css
SP version    : 702
Github commit : 99ed4dca7b439f5fc4ade054355767b23af497df
Server plugins:
   00: Source.Python, (C) 2012-2020, Source.Python Team.
SP plugins:
   00: test
--------------------------------------------------------
```

Not sure if its helpful, but I've used this fix to make Source.Python work:

https://forums.sourcepython.com/viewtopic.php?p=13976#p13976 Method 2

Issue #334, opened 2020-05-13
Linux: @HookUserMessage('SayText2') crashes server


Frag1337:
Hello,

@HookUserMessage('SayText2') crashes on linux when a player is on the server.

```
sp info

IMPORTANT: Please copy the full output.

Checksum      : 0f23eb6e30e633b9a028c12552cb6f25
Date          : 2020-05-13 16:11:47.263541
OS            : Linux-5.4.0-26-generic-x86_64-with-debian-bullseye-sid
Game          : css
SP version    : 695
Github commit : 92b3adfaaaf1bb9ecd2be1ddad55bc68589b4152
Server plugins:
   00: Source.Python, (C) 2012-2019, Source.Python Team.
SP plugins:
   00: test
```

test.py
```
from messages.hooks import HookUserMessage

@HookUserMessage('SayText2')
def on_user_message(recipients, data):
    print(data)
```

Console log:
```
Client "Frag" connected (XXXXXXXXX).
Server waking up from hibernation
Segmentation fault (core dumped)
Add "-debug" to the ./srcds_run command line to generate a debug.log to help with solving this problem
Wed 13 May 2020 06:13:55 PM CEST: Server restart in 10 seconds
```

Frag1337:
Any update?

Ayuto:
Not yet, sorry! I didn't have time to take a look at it.

Ayuto:
Just tested it with the latest SP version (699) and was unable to reproduce it. Could you please check if it was "accidentally" fixed in a later version?
```
Game will not start until both teams have players.
{'index': 1, 'chat': 1, 'message': 'Cstrike_Chat_All', 'param1': 'Ayuto', 'param2': 'asd', 'param3': '', 'param4': ''}
Ayuto: asd

sp info

IMPORTANT: Please copy the full output.
--------------------------------------------------------
Checksum      : a204c8959008c3305614a7de6b86db68
Date          : 2020-08-15 18:11:54.027363
OS            : Linux-4.4.0-144-generic-i686-with-debian-jessie-sid
Game          : css
SP version    : 699
Github commit : 7e06ce8e445e17425dfcecdad5895473c294fabc
Server plugins:
   00: Source.Python, (C) 2012-2020, Source.Python Team.
SP plugins:
   00: test
--------------------------------------------------------
```

Frag1337:
Hey @Ayuto 

Sorry, I sadly had no linux server to test it again.

[...]

PR #375, opened 2020-12-20
Fixed a problem with dangling references around ExtractPointer. (#334, #372, #373)

CookStar:
#334, #372, #373

When the object passed to ExtractPointer isn't the CPointer, the CPointer* generated by _ptr() will be released immediately.
Therefore, the CPointer* generated by ExtractPointer must not be retained.

The new Linux distributions have made this problem more apparent.

### DIFF
--- a/src/core/modules/memory/memory_function.cpp
+++ b/src/core/modules/memory/memory_function.cpp
@@ -280,7 +280,7 @@ object CFunction::Call(tuple args, dict kw)
 			{
 				unsigned long ulAddr = 0;
 				if (arg.ptr() != Py_None)
-					ulAddr = ExtractPointer(arg)->m_ulAddr;
+					ulAddr = ExtractAddress(arg);
 
 				dcArgPointer(g_pCallVM, ulAddr);
 				break;
--- a/src/core/modules/memory/memory_hooks.cpp
+++ b/src/core/modules/memory/memory_hooks.cpp
@@ -147,7 +147,7 @@ bool SP_HookHandler(HookType_t eHookType, CHook* pHook)
 					case DATA_TYPE_DOUBLE:		SetReturnValue<double>(pHook, pyretval); break;
 					case DATA_TYPE_POINTER:
 					{
-						pHook->SetReturnValue<unsigned long>(ExtractPointer(pyretval)->m_ulAddr);
+						pHook->SetReturnValue<unsigned long>(ExtractAddress(pyretval));
 					} break;
 					case DATA_TYPE_STRING:		SetReturnValue<const char*>(pHook, pyretval); break;
 					default: BOOST_RAISE_EXCEPTION(PyExc_TypeError, "Unknown type.")
@@ -224,7 +224,7 @@ void CStackData::SetItem(unsigned int iIndex, object value)
 		case DATA_TYPE_DOUBLE:		SetArgument<double>(m_pHook, iIndex, value); break;
 		case DATA_TYPE_POINTER:
 		{
-			SetArgument<unsigned long>(m_pHook, iIndex, object(ExtractPointer(value)->m_ulAddr));
+			SetArgument<unsigned long>(m_pHook, iIndex, object(ExtractAddress(value)));
 		} break;
 		case DATA_TYPE_STRING:		SetArgument<const char *>(m_pHook, iIndex, value); break;
 		default: BOOST_RAISE_EXCEPTION(PyExc_TypeError, "Unknown type.")
--- a/src/core/modules/memory/memory_hooks.cpp
+++ b/src/core/modules/memory/memory_hooks.cpp
@@ -147,8 +147,7 @@ bool SP_HookHandler(HookType_t eHookType, CHook* pHook)
 					case DATA_TYPE_DOUBLE:		SetReturnValue<double>(pHook, pyretval); break;
 					case DATA_TYPE_POINTER:
 					{
-						CPointer* pPtr = ExtractPointer(pyretval);
-						pHook->SetReturnValue<unsigned long>(pPtr->m_ulAddr);
+						pHook->SetReturnValue<unsigned long>(ExtractPointer(pyretval)->m_ulAddr);
 					} break;
 					case DATA_TYPE_STRING:		SetReturnValue<const char*>(pHook, pyretval); break;
 					default: BOOST_RAISE_EXCEPTION(PyExc_TypeError, "Unknown type.")
@@ -225,8 +224,7 @@ void CStackData::SetItem(unsigned int iIndex, object value)
 		case DATA_TYPE_DOUBLE:		SetArgument<double>(m_pHook, iIndex, value); break;
 		case DATA_TYPE_POINTER:
 		{
-			CPointer* pPtr = ExtractPointer(value);
-			SetArgument<unsigned long>(m_pHook, iIndex, object(pPtr->m_ulAddr));
+			SetArgument<unsigned long>(m_pHook, iIndex, object(ExtractPointer(value)->m_ulAddr));
 		} break;
 		case DATA_TYPE_STRING:		SetArgument<const char *>(m_pHook, iIndex, value); break;
 		default: BOOST_RAISE_EXCEPTION(PyExc_TypeError, "Unknown type.")
--- a/src/core/modules/memory/memory_pointer.cpp
+++ b/src/core/modules/memory/memory_pointer.cpp
@@ -125,8 +125,7 @@ void SetPtrHelper(unsigned long addr, unsigned long ptr)
 void CPointer::SetPtr(object oPtr, int iOffset /* = 0 */)
 {
 	Validate();
-	CPointer* pPtr = ExtractPointer(oPtr);
-	SetPtrHelper(m_ulAddr + iOffset, pPtr->m_ulAddr);
+	SetPtrHelper(m_ulAddr + iOffset, ExtractPointer(oPtr)->m_ulAddr);
 }
 
 int CompareHelper(void* first, void* second, unsigned long length)
@@ -140,18 +139,18 @@ int CompareHelper(void* first, void* second, unsigned long length)
 int CPointer::Compare(object oOther, unsigned long ulNum)
 {
 	Validate();
-	CPointer* pOther = ExtractPointer(oOther);
-	pOther->Validate();
-	return CompareHelper((void *) m_ulAddr, (void *) pOther->m_ulAddr, ulNum);
+	CPointer pOther = CPointer(ExtractPointer(oOther)->m_ulAddr);
+	pOther.Validate();
+	return CompareHelper((void *) m_ulAddr, (void *) pOther.m_ulAddr, ulNum);
 }
 
 bool CPointer::IsOverlapping(object oOther, unsigned long ulNumBytes)
 {
-	CPointer* pOther = ExtractPointer(oOther);
-	if (m_ulAddr <= pOther->m_ulAddr)
-		return m_ulAddr + ulNumBytes > pOther->m_ulAddr;
-       
-	return pOther->m_ulAddr + ulNumBytes > m_ulAddr;
+	unsigned long ulOther = ExtractPointer(oOther)->m_ulAddr;
+	if (m_ulAddr <= ulOther)
+		return m_ulAddr + ulNumBytes > ulOther;
+
+	return ulOther + ulNumBytes > m_ulAddr;
 }
 
 void* SearchBytesHelper(unsigned char* base, unsigned char* end, unsigned char* bytes, unsigned long length)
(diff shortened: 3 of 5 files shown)


jordanbriere:
Nice catch! I can certainly visualize the issue.

Though, unless I missed something obvious, the lifetime of the pointer is now dependant on [oPtr](https://github.com/Source-Python-Dev-Team/Source.Python/blob/4af4b88007df065908ebcaef8e7adab54b2bfef6/src/core/modules/memory/memory_utilities.h#L70), right? If so, I have a feeling the issue will randomly happens again depending if the garbage collector is on lunch break or not at the time it goes out of scope.

Since the pointers instantiated by the Python class are held as shared, and that the inlining would keep the return reference alive throughout the calling scope, wouldn't this be a more robust fix?

```cpp
inline boost::shared_ptr<CPointer> ExtractPointer(object oPtr)
{
	extract<boost::shared_ptr<CPointer> > extractor(oPtr);
	if (!extractor.check())
		return extract<boost::shared_ptr<CPointer> >(oPtr.attr(GET_PTR_NAME)());

	return extractor();
}
```

This should theoretically ensure the pointer's lifetime is extended without relying on a proxy instance. The callers would then simply have to extract the pointer from it like so:

```cpp
boost::get_pointer(ExtractPointer(obj))->m_ulAddr
```

Note that I haven't tested anything and go from a theoretical point of view but if anything, this would prevent the following redundancies. 😆 

```cpp
CPointer pOther = CPointer(ExtractPointer(oOther)->m_ulAddr);
```

CookStar:
> Since the pointers instantiated by the Python class are held as shared, and that the inlining would keep the return reference alive throughout the calling scope, wouldn't this be a more robust fix?
> 
> ```c++
> inline boost::shared_ptr<CPointer> ExtractPointer(object oPtr)
> {
> 	extract<boost::shared_ptr<CPointer> > extractor(oPtr);
> 	if (!extractor.check())
> 		return extract<boost::shared_ptr<CPointer> >(oPtr.attr(GET_PTR_NAME)());
> 
> 	return extractor();
> }
> ```

In my tests, it didn't fix the problem.

Test Code:

```c++
CPointer::~CPointer()
{
	printf("~ CPointer*, %p\n", this);
	printf("~ void *, %p\n", (void*)m_ulAddr);
}
```

```c++
//https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_hooks.cpp#L228
		case DATA_TYPE_POINTER:
		{
			printf("1\n");
			CPointer* pPtr = boost::get_pointer(ExtractPointer(value));
			printf("p %p\n", pPtr);
			printf("v %p\n", (void*)pPtr->m_ulAddr);
			printf("2\n");
			SetArgument<unsigned long>(m_pHook, iIndex, object(pPtr->m_ulAddr));
		} break;
```
Output:
```
1
~ CPointer*, 0xdbb8010
~ void *, 0x99e9f30
p 0xdbb8010
v 0x99e9f30
2

```
It will also double free and crash on exit.


> Though, unless I missed something obvious, the lifetime of the pointer is now dependant on [oPtr](https://github.com/Source-Python-Dev-Team/Source.Python/blob/4af4b88007df065908ebcaef8e7adab54b2bfef6/src/core/modules/memory/memory_utilities.h#L70), right? If so, I have a feeling the issue will randomly happens again depending if the garbage collector is on lunch break or not at the time it goes out of scope.

CPointer is deleted by std::default_delete. If CPointer is generated by _ptr(), it should be released immediately, regardless of the lifetime of oPtr.
I just adjusted the reference before the CPointer is freed.

The behavior in [CPointer::IsOverlapping](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_pointer.cpp#L148).
ida

jordanbriere:
> In my tests, it didn't fix the problem.

Yeah, sorry. That is because `oPtr` wasn't kept alive long enough. But take the following example based on this branch:

```cpp
CPointer::~CPointer()
{
	printf("CPointer::~CPointer(%d)\n", (unsigned long)this);
}

inline CPointer *extract_pointer(object oPtr)
{
	oPtr = oPtr.attr(GET_PTR_NAME)();
	CPointer *pPtr = extract<CPointer *>(oPtr);
	printf("Address of returned pPtr is: %d\n", (unsigned long)pPtr);
	return pPtr;
}

void test_extraction(object oPtr)
{
	CPointer *pPtr = extract_pointer(oPtr);
	printf("Address of extracted pPtr is: %d\n", (unsigned long)pPtr);

	pPtr->m_ulAddr; // Undefined...
}

DECLARE_SP_MODULE(_memory)
{
	def("test_extraction", &test_extraction);
```

```python
import gc
from memory import test_extraction

gc.set_threshold(1, 1, 1)
test_extraction(Vector(1, 2, 3))

"""
Address of returned pPtr is: 443848008
CPointer::~CPointer(443848008)
Address of extracted pPtr is: 443848008
"""
```

The extracted pointer is destroyed before it is accessed because the `CPointer *` returned is bound to and managed by the `PyObject *` returned by `_ptr()` so as soon as `oObj` goes out of scope, it is `decref` and might or might not be collected now, or later when the interpreter needs memory and the thresholds are met, etc.

Now have a look at this example using `shared_ptr`:

```cpp
CPointer::~CPointer()
{
	printf("CPointer::~CPointer(%d)\n", (unsigned long)this);
}

inline boost::shared_ptr<CPointer> extract_shared_pointer(object oPtr)
{
	// Ensure our PyObject is kept alive long enough to incref its shared_ptr...
	oPtr = oPtr.attr(GET_PTR_NAME)();

	// Incref our shared_ptr so we are kept alive no matter if oPtr is collected or not...
	boost::shared_ptr<CPointer> pPtr = extract<boost::shared_ptr<CPointer> >(oPtr);

	printf("Address of returned pPtr is: %d\n", (unsigned long)boost::get_pointer(pPtr));
	return pPtr;
}

void test_shared_extraction(object oPtr)
{
	// Our shared_ptr will live throughout this scope...
	boost::shared_ptr<CPointer> pPtr = extract_shared_pointer(oPtr);

	printf("Address of extracted pPtr is: %d\n", (unsigned long)boost::get_pointer(pPtr));

	boost::get_pointer(pPtr)->m_ulAddr; // No longer undefined...
}

DECLARE_SP_MODULE(_memory)
{
	def("test_shared_extraction", &test_shared_extraction);
```

```python
import gc
from memory import test_shared_extraction

gc.set_threshold(1, 1, 1)
test_shared_extraction(Vector(1, 2, 3))

"""
Address of returned pPtr is: 430348360
Address of extracted pPtr is: 430348360
CPointer::~CPointer(430348360)
"""
```

So yeah, by adding `oPtr = oPtr.attr(GET_PTR_NAME)();`, you increased the lifetime of `CPointer *` long enough so it is valid at the time of extracting (which I guess is where the crashes were happening) but accessing the pointer once returned is still undefined and may randomly cause other crashes if the interpreter re-used that memory already.

Frag1337:
Is there any date when this is gonna be merged? My linux servers would be very happy about this.

CookStar:
> Yeah, sorry. That is because `oPtr` wasn't kept alive long enough.
I also forgot to hold the boost::shared_ptr<CPointer> object in my code.

> The extracted pointer is destroyed before it is accessed because the `CPointer *` returned is bound to and managed by the `PyObject *` returned by `_ptr()` so as soon as `oObj` goes out of scope, it is `decref` and might or might not be collected now, or later when the interpreter needs memory and the thresholds are met, etc.

This is true if we return `CPointer *`, however, what we are actually returning is `extract<CPointer *>(oPtr)`, so as long as we refer to rvalue directly, oPtr will not be destroyed.

```c++
//https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/thirdparty/boost/boost/python/object_core.hpp#L245
//	  ~object(){
//	  	printf("~object, %p\n", this);
//	  }


inline CPointer* extract_c_pointer(object oPtr)
{
	extract<CPointer *> extractor(oPtr);
	if (!extractor.check()){
		oPtr = oPtr.attr(GET_PTR_NAME)();
		printf("oPtr*, %p\n", &oPtr);
		return extract<CPointer *>(oPtr);
	}

	return extractor();
}

void test(object oPtr)
{
	printf("1\n");
	printf("void*, %p\n", (void*)(extract_c_pointer(oPtr)->m_ulAddr));
	printf("2\n");
}
```
Output:
```
1
~object, 0xff948980
~object, 0xff948960
~object, 0xff948998
oPtr*, 0xff948940
void*, 0xf018b250
~object, 0xff948940
2
```

If ExtractPointer were a more universally used function, shared_ptr would be better, but it is actually used in only 9 places, and even if we created a CPointer, it is still 20% faster than shared_ptr. It is wasteful to create an unnecessary shared_ptr just to reference the memory address of c objects.
I think it's better to leave the ExtractPointer as it is for optimization.

If you disagree, I would change it to shared_ptr.

jordanbriere:
> This is true if we return `CPointer *`, however, what we are actually returning is `extract<CPointer *>(oPtr)`, so as long as we refer to rvalue directly, oPtr will not be destroyed.

Well, `extract<CPointer *>(oPtr)` is really just a call that returns a `CPointer *` and doesn't hold any reference whatsoever once returned. The only thing I could see having an impact on the longevity of `oPtr`, is some compiler-specific optimizations due to inlining/rewriting an existing reference. But even then, it isn't guaranteed to stay alive and would likely be on a case-by-case basis whether it is considered volatile or not (e.g. extraction is made within a loop, forwarded to a nested call, etc.).

I reproduced your test just now:

```cpp
object::~object()
{
	printf("object::~object(%d)\n", this);
}

CPointer::~CPointer()
{
	printf("CPointer::~CPointer(%d)\n", this);
}

unsigned long CPointer::GetAddress()
{
	printf("CPointer::GetAddress(%d)\n", this);
	return m_ulAddr;
}

inline CPointer* extract_c_pointer(object oPtr)
{
	extract<CPointer *> extractor(oPtr);
	if (!extractor.check()){
		oPtr = oPtr.attr(GET_PTR_NAME)();
		printf("oPtr*, %d\n", &oPtr);
		return extract<CPointer *>(oPtr);
	}

	return extractor();
}

void test(object oPtr)
{
	printf("1\n");
	printf("void*, %d\n", (void*)(extract_c_pointer(oPtr)->m_ulAddr));
	printf("2\n");
}

DECLARE_SP_MODULE(_memory)
{
	def("test", &test);
```

```python
test(Vector(1, 2, 3))

"""
1
object::~object(3791940)
object::~object(3791992)
object::~object(3791984)
oPtr*, 3792016
object::~object(3792016)
CPointer::~CPointer(431134080)
void*, 444460984
2
"""
```

Once `ExtractPointer` returns, the fate of `oPtr`, along with the `CPointer *` it manages, is undefined. So is accessing any of its members/calling any of its method.


> If ExtractPointer were a more universally used function, shared_ptr would be better, but it is actually used in only 9 places, and even if we created a CPointer, it is still 20% faster than shared_ptr. It is wasteful to create an unnecessary shared_ptr just to reference the memory address of c objects.
> I think it's better to leave the ExtractPointer as it is for optimization.

Technically, this is not creating a `shared_ptr`. The `Pointer` class already manage its object as such, so extracting it as shared really just return an existing reference and incref it.

> If you disagree, I would change it to shared_ptr.

To be honest, I have issues with both. Returning a `CPointer *` is prone to issues, just as much as returning a `shared_ptr` is unless it is specifically held within the calling scope. I think I would overall prefer... none of them. If we look at the current usages, all calls really just have a need for `m_ulAddr` and the only thing that differs, is whether it is validated or not. I think I would rather prefer and suggest to replace the current `ExtractPointer` with the following:

```cpp
inline unsigned long ExtractAddress(object oPtr, bool bValidate = false)
{
	extract<CPointer *> extractor(oPtr);
	if (!extractor.check())
		oPtr = oPtr.attr(GET_PTR_NAME)();

	CPointer *pPtr = extract<CPointer *>(oPtr);
	if (bValidate)
		pPtr->Validate();

	return pPtr->m_ulAddr;
}
```

This would ensure the pointer is still alive when accessed, as well as removing any reference mess the other alternatives are suffering.

CookStar:
> Well, `extract<CPointer *>(oPtr)` is really just a call that returns a `CPointer *` and doesn't hold any reference whatsoever once returned. The only thing I could see having an impact on the longevity of `oPtr`, is some compiler-specific optimizations due to inlining/rewriting an existing reference. But even then, it isn't guaranteed to stay alive and would likely be on a case-by-case basis whether it is considered volatile or not (e.g. extraction is made within a loop, forwarded to a nested call, etc.).

> Once `ExtractPointer` returns, the fate of `oPtr`, along with the `CPointer *` it manages, is undefined. So is accessing any of its members/calling any of its method.

> Technically, this is not creating a `shared_ptr`. The `Pointer` class already manage its object as such, so extracting it as shared really just return an existing reference and incref it.

You are absolutely correct, my understanding was incomplete.

> To be honest, I have issues with both. Returning a `CPointer *` is prone to issues, just as much as returning a `shared_ptr` is unless it is specifically held within the calling scope. I think I would overall prefer... none of them. If we look at the current usages, all calls really just have a need for `m_ulAddr` and the only thing that differs, is whether it is validated or not. I think I would rather prefer and suggest to replace the current `ExtractPointer` with the following:
> 
> ```c++
> inline unsigned long ExtractAddress(object oPtr, bool bValidate = false)
> {
> 	extract<CPointer *> extractor(oPtr);
> 	if (!extractor.check())
> 		oPtr = oPtr.attr(GET_PTR_NAME)();
> 
> 	CPointer *pPtr = extract<CPointer *>(oPtr);
> 	if (bValidate)
> 		pPtr->Validate();
> 
> 	return pPtr->m_ulAddr;
> }
> ```
> 
> This would ensure the pointer is still alive when accessed, as well as removing any reference mess the other alternatives are suffering.

I was considering that too, but I gave up after thinking about MakeObject, apparently I was overthinking about that.
Made some small adjustments related to extractor and committed the fix!

jordanbriere (approved):

PR #376, opened 2020-12-21
Fixed compilation errors with CS:GO and Blade.

CookStar:
Fixed `numerous errors` when including "player.h" and multiple definitions. #363


Ayuto:
Thanks! I have added the fix for the multiple defintions, but chose a different approach to fix the include errors, because I don't want to addd a patch for that.

PR #380, opened 2021-01-23
Added CEnvProjectedTexture data.

vinci6k:


satoon101:
I know this was already merged, but I have a couple points to add.  I just tested the output on CS:S, HL2:MP, DOD:S, TF2, and CS:GO.  I found that a few of the inputs, keyvalues, and properties ONLY exist on CS:GO:
* Inputs
  * AlwaysUpdateOff
  * AlwaysUpdateOn
* Keyvalues
  * brightnessscale
  * colortransitiontime
  * projection_rotation
  * projection_size
  * simpleprojection
* Properities
  * m_bAlwaysUpdate
  * m_LightColor

Also, `texturename` is a Property for all games, not a Keyvalue.

For the orangebox games, they have a Property `m_LinearFloatLightColor` that doesn't exist in CS:GO (which 'might' be the same as m_LightColor of CS:GO, but I'm not sure).

The code I used to find these was:
```python

entity = Entity.find_or_create('env_projectedtexture')

for k, server_class in entity.server_classes.items():
    if k.__name__ == 'CEnvProjectedTexture':
        break

for type_name in ('Inputs', 'Keyvalues', 'Outputs', 'Properties'):
    print('=' * 30)
    print(type_name.upper().center(30))
    print('=' * 30)
    for item in sorted(getattr(server_class, type_name.lower())):
        print(f'    {item}')
```

vinci6k:
Whoops, my bad. I'll correct this with another pull request as soon as I get home.
Although m_LightColor seems to work in HL2DM - at least it did when I tried to port [flashlight.py](https://forums.sourcepython.com/viewtopic.php?f=37&t=2467#p15144) to it.

jordanbriere:
> Whoops, my bad. I'll correct this with another pull request as soon as I get home.

Arggh.. I trusted you, I TRUSTED YOU!! 😆 

> Although m_LightColor seems to work in HL2DM - at least it did when I tried to port [flashlight.py](https://forums.sourcepython.com/viewtopic.php?f=37&t=2467#p15144) to it.

Sometimes keyvalues are mapped to different members depending on the game. Some are also hard-coded and processed/validated in various ways before assigning/retrieving them. This seems to be the case for [lightcolor](https://github.com/alliedmodders/hl2sdk/blob/hl2dm/game/server/env_projectedtexture.cpp#L158) on OB games as well as pointing to a different property.

PR #381, opened 2021-01-25
Split CEnvProjectedTexture data between different games.

vinci6k:
###### Continuation of #380 - due to me not testing the data properly (sorry @jordanbriere :sweat_smile:).
Tested the changes in HL2DM, CSS, and CSGO.
The only issue I ran into was the hardcoded [lightcolor](https://github.com/alliedmodders/hl2sdk/blob/hl2dm/game/server/env_projectedtexture.cpp#L158) keyvalue - setting the type as Color makes the data unusable: 
```
[SP] Caught an Exception:
  File "..\addons\source-python\packages\source-python\commands\auth.py", line 44, in __call__
    return self.callback(*args)
  File "..\addons\source-python\plugins\flashlight\flashlight.py", line 52, in inspect_weapon
    PlayerF(index).toggle_torch()
  File "..\addons\source-python\plugins\flashlight\flashlight.py", line 111, in toggle_torch
    owner_handle=self.inthandle
  File "..\addons\source-python\plugins\flashlight\flashlight.py", line 157, in create
    texture.light_color = color
  File "..\addons\source-python\packages\source-python\entities\_base.py", line 246, in __setattr__
    for server_class, instance in self.server_classes.items():
  File "..\addons\source-python\packages\source-python\entities\_base.py", line 231, in __getattr__
    raise AttributeError('Attribute "{0}" not found'.format(attr))

AttributeError: Attribute "server_classes" not found
```
This is quite odd as setting the same keyvalue with set_key_value_color() works fine. What am I doing wrong here?
For the time being, I renamed that keyvalue to light_color_string and set the type as STRING.

@satoon101 was correct about `m_LinearFloatLightColor` being the Orange Box/L4D2 equivalent of `m_LightColor` from CSGO, but it doesn't use the Color object. Instead it uses a Vector containing floats ranging from 0 to 1 (e.g. Vector(1, 0, 0) would be red).

### DIFF
--- a/addons/source-python/data/source-python/entities/CEnvProjectedTexture.ini
+++ b/addons/source-python/data/source-python/entities/CEnvProjectedTexture.ini
@@ -1,7 +1,5 @@
 [input]
 
-    always_update_off = AlwaysUpdateOff
-    always_update_on = AlwaysUpdateOn
     fov = FOV
     spotlight_texture = SpotlightTexture
     target = Target
@@ -12,29 +10,22 @@
 [keyvalue]
     
     ambient = ambient
-    brightness_scale = brightnessscale
     camera_space = cameraspace
-    color_transition_time = colortransitiontime
     enable_shadows = enableshadows
     far_z = farz
     light_fov = lightfov
     light_only_target = lightonlytarget
     light_world = lightworld
     near_z = nearz
-    projection_rotation = projection_rotation
-    projection_size = projection_size
     shadow_quality = shadowquality
-    simple_projection = simpleprojection
     texture_frame = textureframe
-    texture_name = texturename
 
+    [[light_color_string]]
+        name = lightcolor
+        type = STRING
 
 [property]
     
-    always_update = m_bAlwaysUpdate
     state = m_bState
     target_entity = m_hTargetEntity
-
-    [[light_color]]
-        name = m_LightColor
-        type = Color
+    texture_name = texturename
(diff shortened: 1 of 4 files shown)


jordanbriere:
> The only issue I ran into was the hardcoded lightcolor keyvalue - setting the type as Color makes the data unusable:

The error you get isn't quite explicit, but I believe the `AttributeError` caught is in fact when the manager attempts to get `FieldType.Color` which doesn't exists. To map an hardcoded keyvalue to `<get/set>_key_value_color`, you have to use [COLOR32](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/entities/classes.py#L82) as type. However, since it appears to be translated as a string of floats, the vector methods are likely to raise about it if the values are between 0-1 because they expect a string of ints so defining it as string is probably for the best anyway.

vinci6k:
> To map an hardcoded keyvalue to `<get/set>_key_value_color`, you have to use [COLOR32](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/entities/classes.py#L82) as type.

Ah.. so that was the problem.

> However, since it appears to be translated as a string of floats, the color methods are likely to raise about it if the values are between 0-1 because they expect a string of ints so defining it as string is probably for the best anyway.


Using `set_key_value_color('lightcolor', color)` works, but trying to use `get_key_value_color('lightcolor')` raises:
```
ValueError: KeyValue does not seem to be a color: ''.
```
Guess I'll leave it as a string then.

PR #383, opened 2021-02-01
Added ShowMenu implementation

vinci6k:
###### As promised: https://forums.sourcepython.com/viewtopic.php?f=37&t=2472#p15210

Tested in CSS and CSGO with:
```py
# ../showmenu_hook/showmenu_hook.py

# Source.Python
from messages.hooks import HookUserMessage


@HookUserMessage('ShowMenu')
def showmenu_pre(recipients, buffer):
    print(recipients, buffer)
```

### DIFF
--- a/addons/source-python/packages/source-python/messages/impl.py
+++ b/addons/source-python/packages/source-python/messages/impl.py
@@ -194,6 +194,7 @@ class ShowMenuImpl(UserMessageImpl):
         return UserMessageData(
             valid_slots=buffer.read_word(),
             display_time=buffer.read_char(),
+            chunked=buffer.read_byte(),
             menu_string=buffer.read_string()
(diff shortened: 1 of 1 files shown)


Ayuto:
Thank you! Have you check very big menus on CS:S (bitbuffer)? Because there should actually be a byte that tells us whether or not the menu is chunked (right before the menu string):
https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/packages/source-python/messages/base.py#L242-L263

vinci6k:
I used [change_map.py](https://forums.sourcepython.com/viewtopic.php?f=37&t=2415&p=14826#p14826) to test the hook and oddly enough, nothing seemed wrong. 😅

PR #387, opened 2021-02-11
Fixed referencing of released m_pCallingConvention on Source.Python unload.

CookStar:
UnhookAllFunctions releases all m_pCallingConventions, but CFunctions held by AutoUnload such as EntityPreHook are not yet released, causing segmentation fault.
https://github.com/Source-Python-Dev-Team/Source.Python/blob/89273cd840e601094f31ade3ef601ce3a95f012c/src/core/modules/memory/memory_function.cpp#L199
UnhookAllFunctions has been moved to the end to deal with this.


jordanbriere:
Thanks, I was able to reproduce (not the crash, but the behaviours which I can see randomly leading to one). This also fixes potential issues if a plugin was dumb enough to register hooks in its `unload` method.

PR #396, opened 2021-05-13
Fixed MemberFunction not returning the correct return type.

CookStar:
If the Function has a converter, then the MemberFunction must also have a correct converter.

### DIFF
--- a/src/core/modules/memory/memory_function.cpp
+++ b/src/core/modules/memory/memory_function.cpp
@@ -179,6 +179,29 @@ CFunction::CFunction(unsigned long ulAddr, Convention_t eCallingConvention,
 	m_oConverter = oConverter;
 }
 
+CFunction::CFunction(const CFunction& obj)
+	:CPointer(obj)
+{
+	m_tArgs = obj.m_tArgs;
+	m_eReturnType = obj.m_eReturnType;
+	m_oConverter = obj.m_oConverter;
+
+	m_eCallingConvention = obj.m_eCallingConvention;
+	m_iCallingConvention = obj.m_iCallingConvention;
+
+	if (m_eCallingConvention != CONV_CUSTOM)
+	{
+		m_pCallingConvention = MakeDynamicHooksConvention(m_eCallingConvention, ObjectToDataTypeVector(m_tArgs), m_eReturnType);
+		m_oCallingConvention = object();
+	}
+	else
+	{
+		m_pCallingConvention = obj.m_pCallingConvention;
+		m_oCallingConvention = obj.m_oCallingConvention;
+		Py_INCREF(m_oCallingConvention.ptr());
+	}
+}
+
 CFunction::~CFunction()
 {
 	if (!m_pCallingConvention)
--- a/src/core/modules/memory/memory_function.h
+++ b/src/core/modules/memory/memory_function.h
@@ -53,12 +53,14 @@ enum Convention_t
 // ============================================================================
 // >> CFunction
 // ============================================================================
-class CFunction: public CPointer, private boost::noncopyable
+class CFunction: public CPointer
 {
 public:
 	CFunction(unsigned long ulAddr, object oCallingConvention, object oArgs, object oReturnType);
 	CFunction(unsigned long ulAddr, Convention_t eCallingConvention, int iCallingConvention,
 		boost::python::tuple tArgs, DataType_t eReturnType, object oConverter);
+
+	CFunction(const CFunction& obj);
 
 	~CFunction();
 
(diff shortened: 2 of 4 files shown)


jordanbriere:
> 
> 
> If the Function has a converter, then the MemberFunction must also have a correct converter.

Thank you! This is a regression that was introduced when [removing the copy constructor of the `Function` class](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_wrap.cpp#L457-L458) to prevent reference/leak issues with its convention. Now that all (most?) convention issues have been addressed perhaps we could look into adding it back instead?

CookStar:
> Now that all (most?) convention issues have been addressed perhaps we could look into adding it back instead?

Indeed, I don't think there will be any problems, either.
Restored the copy constructor.

One question: In this [part added](https://github.com/CookStar/Source.Python/blob/6d26bca14429b498e5862ccbbfbae57fd0bb7c8a/src/core/modules/memory/memory_function.cpp#L199-L200) by https://github.com/Source-Python-Dev-Team/Source.Python/pull/344/commits/6d26bca14429b498e5862ccbbfbae57fd0bb7c8a, why does it check `m_pCallingConvention` by dynamic_cast and not just by comparing with `CONV_CUSTOM`?
https://github.com/Source-Python-Dev-Team/Source.Python/blob/227e9546b395cbf5f9d4a348363b8eebb1bd527e/src/core/modules/memory/memory_function.cpp#L199-L200
If `m_eCallingConvention` isn't `CONV_CUSTOM`, I suppose there's no need to check further, is there?

src/core/modules/memory/memory_function.h
```diff
@@ -105,7 +107,6 @@ class CFunction: public CPointer, private boost::noncopyable
 
 	// DynamicHooks calling convention (built-in and custom)
 	ICallingConvention*		m_pCallingConvention;
-	bool					m_bAllocatedCallingConvention;
```
CookStar, line 110: Removed m_bAllocatedCallingConvention, which was forgotten.

src/core/modules/memory/memory_function.cpp
```diff
@@ -179,6 +179,29 @@ CFunction::CFunction(unsigned long ulAddr, Convention_t eCallingConvention,
...
+	if (m_eCallingConvention != CONV_CUSTOM)
+	{
+		m_pCallingConvention = MakeDynamicHooksConvention(m_eCallingConvention, ObjectToDataTypeVector(m_tArgs), m_eReturnType);
+		m_oCallingConvention = object();
+	}
+	else
+	{
+		m_pCallingConvention = obj.m_pCallingConvention;
+		m_oCallingConvention = obj.m_oCallingConvention;
+		Py_INCREF(m_oCallingConvention.ptr());
```
jordanbriere, line 201: I haven't tested, only hovered the changes, but is that necessary? I guess this is an additional reference reserved for DynamicHooks and it will be taken care of as well when the copy gets destroyed?

jordanbriere (commented):
> One question: In this [part added](https://github.com/CookStar/Source.Python/blob/6d26bca14429b498e5862ccbbfbae57fd0bb7c8a/src/core/modules/memory/memory_function.cpp#L199-L200) by [6d26bca](https://github.com/Source-Python-Dev-Team/Source.Python/commit/6d26bca14429b498e5862ccbbfbae57fd0bb7c8a), why does it check `m_pCallingConvention` by dynamic_cast and not just by comparing with `CONV_CUSTOM`?

It was added as extra security as mentioned into https://github.com/Source-Python-Dev-Team/Source.Python/pull/344#discussion_r500581299. However, feel free to remove it. Chances are that whole thing will collapse anyway if they are independently mutated. 😄

jordanbriere:
Had a chance to run some tests just now and it passed everything I threw at it. Well done!

CookStar:
I have been unable to access github for a long time, so I apologize for the delay in replying.

> It was added as extra security as mentioned into [#344 (comment)](https://github.com/Source-Python-Dev-Team/Source.Python/pull/344#discussion_r500581299). However, feel free to remove it. Chances are that whole thing will collapse anyway if they are independently mutated. 😄

Indeed, you were right, it totally dropped from my memory.

src/core/modules/memory/memory_function.cpp
CookStar: This process is necessary to prevent double freeing in the destructor, but it seems to cause a memory leak when the function is hooked related to another bug.

Details will be provided in a new pull request.

Issue #390, opened 2021-03-26
Menu exception 


NosferatuJoe:
I encountered an exception with the menus, not sure if it's a big issue but thought to report it nevertheless.

```
2021-03-26 19:48:01 - sp	-	EXCEPTION	
[SP] Caught an Exception:
Traceback (most recent call last):
  File "../addons/source-python/packages/source-python/commands/auth.py", line 44, in __call__
    return self.callback(*args)
  File "../addons/source-python/packages/source-python/menus/queue.py", line 267, in _menuselect_callback
    _radio_queues[index]._select(choice)
  File "../addons/source-python/packages/source-python/menus/queue.py", line 126, in _select
    next_menu = active_menu._select(self._index, choice)
  File "../addons/source-python/packages/source-python/menus/radio.py", line 113, in _select
    self._player_pages[player_index].options[choice_index])

KeyError: 1
```

I might've been using the adaptive radar plugin from the forums though, not sure if it was turned on or off.

jordanbriere:
Could you please provide a reproducible code? That traceback is rather ambiguous, as there is no way to tell if the missing key is the player index or the selected option..

vinci6k:
He mentioned he was might've been using the [adaptive radar](https://forums.sourcepython.com/viewtopic.php?f=37&t=2472#p15224) plugin, so I have a feeling this might be causing issues with menus created by SourceMod plugins:
```py
@ClientCommand('menuselect')
def menu_select(command, index):
    """Called when the player selects an option in the menu."""
    # Did the player close their menu?
    if SimpleMenu.get_user_queue(index).active_menu is None:
        PlayerAR(index).show_radar()
```
I'll see if I can reproduce it in a few minutes.

vinci6k:
Ran into a small issue:
```
[SP] Caught an Exception:
Traceback (most recent call last):
  File "..\addons\source-python\packages\source-python\messages\hooks.py", line 188, in _pre_send_user_message
    buffer = make_object(ProtobufMessage, args[3])

RuntimeError: Access violation - no RTTI data!
```
That happens whenever a menu is created/sent by SourceMod.

jordanbriere:
> 
> 
> Ran into a small issue:
> 
> ```
> [SP] Caught an Exception:
> Traceback (most recent call last):
>   File "..\addons\source-python\packages\source-python\messages\hooks.py", line 188, in _pre_send_user_message
[...]

PR #399, opened 2021-05-22
Partial fix for issue #390.

vinci6k:
This fixes the hooks from not getting called when the UserMessage was created/sent by a third-party plugin (SourceMod), but I'm not sure how to retrieve the actual UserMessage data (e.g. [ShowMenu data](https://github.com/Source-Python-Dev-Team/Source.Python/blob/35d34c885b2d787b265c99930d0fe09f5534668d/addons/source-python/packages/source-python/messages/impl.py#L181)).

### DIFF
--- a/addons/source-python/packages/source-python/messages/hooks.py
+++ b/addons/source-python/packages/source-python/messages/hooks.py
@@ -186,14 +186,7 @@ if UserMessage.is_protobuf():
             buffer = make_object(ProtobufMessage, args[3])
         except RuntimeError:
             # Patch for issue #390 - UserMessage was created by another plugin.
-            buffer = UserMessage(
-                _recipients, get_message_name(message_index)).buffer
-            buffer_ptr = get_object_pointer(buffer)
-            buffer_size = get_size(buffer)
-
-            orig_buffer_ptr = args[3]
-            orig_buffer_ptr.copy(buffer_ptr, buffer_size)
-            buffer_ptr.copy(orig_buffer_ptr, buffer_size)
+            buffer = ProtobufMessage.from_abstract_pointer(args[3])
 
         protobuf_user_message_hooks.notify(_recipients, buffer)
 
--- a/addons/source-python/packages/source-python/messages/hooks.py
+++ b/addons/source-python/packages/source-python/messages/hooks.py
@@ -182,7 +182,18 @@ if UserMessage.is_protobuf():
         _recipients.update(*tuple(tmp_recipients), clear=True)
         args[1] = _recipients
 
-        buffer = make_object(ProtobufMessage, args[3])
+        try:
+            buffer = make_object(ProtobufMessage, args[3])
+        except RuntimeError:
+            # Patch for issue #390 - UserMessage was created by another plugin.
+            buffer = UserMessage(
+                _recipients, get_message_name(message_index)).buffer
+            buffer_ptr = get_object_pointer(buffer)
+            buffer_size = get_size(buffer)
+
+            orig_buffer_ptr = args[3]
+            orig_buffer_ptr.copy(buffer_ptr, buffer_size)
+            buffer_ptr.copy(orig_buffer_ptr, buffer_size)
 
         protobuf_user_message_hooks.notify(_recipients, buffer)
 


jordanbriere:
Nice one! What do you mean by retrieving the data?

vinci6k:
Sorry for not being more clear. The following code should print the contents of a sent menu:
```py
from messages.hooks import HookProtobufUserMessage


@HookProtobufUserMessage('ShowMenu')
def show_menu_pre(recipients, buffer):
    """Called when a ShowMenu user message gets sent."""
    print(buffer.get_string('menu_string'))
```
It does with SP menus, but not with SM menus. 
I understand that by creating a new UserMessage, the buffer will be [empty](https://github.com/Source-Python-Dev-Team/Source.Python/blob/35d34c885b2d787b265c99930d0fe09f5534668d/src/core/modules/messages/messages.cpp#L60). So I'm looking for a way to somehow get that data, if such a thing is even possible?

jordanbriere:
> 
> 
> Sorry for not being more clear. The following code should print the contents of a sent menu:
> 
> ```python
> from messages.hooks import HookProtobufUserMessage
> 
> 
> @HookProtobufUserMessage('ShowMenu')
> def show_menu_pre(recipients, buffer):
>     """Called when a ShowMenu user message gets sent."""
>     print(buffer.get_string('menu_string'))
> ```
> 
> It does with SP menus, but not with SM menus.
> I understand that by creating a new UserMessage, the buffer will be [empty](https://github.com/Source-Python-Dev-Team/Source.Python/blob/35d34c885b2d787b265c99930d0fe09f5534668d/src/core/modules/messages/messages.cpp#L60). So I'm looking for a way to somehow get that data, if such a thing is even possible?

Ah yeah. There are two possible causes;

- Its size is dynamic and not enough data is copied because `get_size` returns the size of the interface, not the actual message.
- The calls aren't mapped properly because the dispatch table was overwritten. Not copying the first 4 bytes could work.

That said, the ideal solution would be to simply have a static constructor (for example, like [RecipientFilter.from_abstract](https://github.com/Source-Python-Dev-Team/Source.Python/commit/86f57c550e274006e7380dbfd8cd682dfc90eb3e) that create a new message and make a copy of the fields using [Message::CopyFrom](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.CopyFrom.details) -- either that or export those directly though more efficient through a one-call constructor.

vinci6k:
Yep, it's the size. As soon as I added a few bytes it started working properly:
```py
buffer_size = get_size(buffer) + 16
```
Running this and opening the admin menu from SourceMod:
```py
@HookProtobufUserMessage('ShowMenu')
def show_menu_pre(recipients, buffer):
    """Called when a ShowMenu user message gets sent."""
    print(buffer.get_int32('bits_valid_slots'))
    print(buffer.get_int32('display_time'))
    print(buffer.get_string('menu_string'))
```
Prints this:
```
263
0
Admin Menu:
->1. Player Commands
->2. Server Commands
->3. Voting Commands

->9. Exit
```




> That said, the ideal solution would be to simply have a static constructor (for example, like [RecipientFilter.from_abstract](https://github.com/Source-Python-Dev-Team/Source.Python/commit/86f57c550e274006e7380dbfd8cd682dfc90eb3e) that create a new message and make a copy of the fields using [Message::CopyFrom](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.CopyFrom.details) -- either that or export those directly though more efficient through a one-call constructor.

I'd love to do that, but I'm kinda out of my depth when it comes to stuff like this. 😅

jordanbriere:
Untested in a real scenario, but try this: [core.dll](https://drive.google.com/drive/folders/1mExLrpfxqrn4Glcv0iw_oFTWoTnUee_v?usp=sharing)

Here's the basic test I ran:

```python
from _messages import ProtobufMessage
from messages import UserMessage
from filters.recipients import RecipientFilter

message = UserMessage(RecipientFilter(), 'ShowMenu')
buffer = message.buffer
buffer.set_string('menu_string', 'blah blah blah')
copy = ProtobufMessage.from_abstract_pointer(buffer)
assert buffer.debug_string == copy.debug_string

```

vinci6k:
Tried this (going by how it's done [here](https://github.com/Source-Python-Dev-Team/Source.Python/commit/86f57c550e274006e7380dbfd8cd682dfc90eb3e)) inside message/hooks.py:
```py
buffer = ProtobufMessage.from_abstract_pointer(args[3])
```
And I got this:
```
[SP] Caught an Exception:
  File "..\addons\source-python\packages\source-python\messages\hooks.py", line 191, in _pre_send_user_message
    buffer = make_object(ProtobufMessage, args[3])
RuntimeError: Access violation - no RTTI data!

During handling of the above exception, another exception occurred:
  File "..\addons\source-python\packages\source-python\messages\hooks.py", line 199, in _pre_send_user_message
    buffer = ProtobufMessage.from_abstract_pointer(args[3])

RuntimeError: Access violation - no RTTI data!
```

jordanbriere:
Hmm, that's weird. Updated the build above, what's the last number you get?

vinci6k:
Seems to be 3:
```
1
2
3

[SP] Caught an Exception:
  File "..\addons\source-python\packages\source-python\messages\hooks.py", line 191, in _pre_send_user_message
    buffer = make_object(ProtobufMessage, args[3])
RuntimeError: Access violation - no RTTI data!

During handling of the above exception, another exception occurred:
  File "..\addons\source-python\packages\source-python\messages\hooks.py", line 199, in _pre_send_user_message
    buffer = ProtobufMessage.from_abstract_pointer(args[3])

RuntimeError: Access violation - no RTTI data!
```

jordanbriere:
Then that means `Message::MergeFrom` itself  internally does some dynamic type checking. Try the updated build above.

vinci6k:
It works perfectly!
```
263
0
Admin Menu:
->1. Player Commands
->2. Server Commands
->3. Voting Commands

->9. Exit
```

addons/source-python/packages/source-python/messages/hooks.py
```diff
@@ -182,7 +182,11 @@ def _pre_send_user_message(args):
         _recipients.update(*tuple(tmp_recipients), clear=True)
         args[1] = _recipients
 
-        buffer = make_object(ProtobufMessage, args[3])
+        try:
+            buffer = make_object(ProtobufMessage, args[3])
+        except RuntimeError:
+            # Patch for issue #390 - UserMessage was created by another plugin.
+            buffer = ProtobufMessage.from_abstract_pointer(args[3])
```
jordanbriere, line 189: The problem though, is that any changes made to the `buffer`'s fields within a plugin's callback will have no effect because `args[3]` is never mutated. We could simply overwrite it with our own, but that is likely to cause some scoping issues since unlike bitbuf messages, I believe protobuf ones can be nested, etc. I went ahead and added `ProtobufMessage.parse_to_abstract_pointer` so that we can ensure any changes that were made is properly applied to the original buffer. For example, something like this (untested):

```python
        # No need to do anything behind this if no listener is registered
        if not user_message_hooks and not protobuf_user_message_hooks:
            return

        try:
            # Replace original recipients filter
            tmp_recipients = make_object(BaseRecipientFilter, args[1])
        except RuntimeError:
            # Patch for issue #314
            tmp_recipients = RecipientFilter.from_abstract_pointer(args[1])
        _recipients.update(*tuple(tmp_recipients), clear=True)
        args[1] = _recipients

        try:
            buffer = make_object(ProtobufMessage, args[3])
        except RuntimeError:
            # Patch for issue #390 - UserMessage was created by another plugin.
            buffer = ProtobufMessage.from_abstract_pointer(args[3])

        protobuf_user_message_hooks.notify(_recipients, buffer)

        # No need to do anything behind this if no listener is registered
        if user_message_hooks:
            try:
                impl = get_user_message_impl(message_index)
            except NotImplementedError:
                impl = None

            if impl is not None:
                data = impl.read(buffer)
                user_message_hooks.notify(_recipients, data)

                # Update buffer if data has been changed
                if data.has_been_changed():
                    buffer.clear()
                    impl.write(buffer, data)

        # If we wrapped the buffer from an abstract pointer, make sure to
        #   apply any changes that may have been made back into the original.
        if get_object_pointer(buffer) != args[3]:
            buffer.parse_to_abstract_pointer(args[3])
```

If this works for you, please update this PR and adjust if needs be. Thanks!

PR #414, opened 2021-09-22
Moved on_convar_changed_listener_manager to C++.

CookStar:
Improved performance and eliminated hooks.

As a backwards compatibility break, when `CallGlobalChangeCallbacks` is blocked, the listener will no longer be emitted, but I don't think this will have any impact.

If there is a problem, I will reset to the hook.

### DIFF
--- a/src/core/modules/listeners/listeners_manager.cpp
+++ b/src/core/modules/listeners/listeners_manager.cpp
@@ -173,7 +173,7 @@ void CListenerManager::clear()
 //-----------------------------------------------------------------------------
 void ConVarChangedCallback(IConVar* var, const char* pOldValue, float flOldValue)
 {
-	CConVarChangedListenerManager* on_convar_changed_listener_manager = GetOnConVarChangedListenerManager();
+	static CConVarChangedListenerManager* on_convar_changed_listener_manager = GetOnConVarChangedListenerManager();
 	CALL_LISTENERS_WITH_MNGR(on_convar_changed_listener_manager, ptr(static_cast<ConVar*>(var)), pOldValue);
 }
 
--- a/src/core/modules/listeners/listeners_manager.cpp
+++ b/src/core/modules/listeners/listeners_manager.cpp
@@ -31,6 +31,18 @@
 
 
 //-----------------------------------------------------------------------------
+// External variables.
+//-----------------------------------------------------------------------------
+extern ICvar* g_pCVar;
+
+
+//-----------------------------------------------------------------------------
+// Extern functions
+//-----------------------------------------------------------------------------
+extern CConVarChangedListenerManager* GetOnConVarChangedListenerManager();
+
+
+//-----------------------------------------------------------------------------
 // Adds a callable to the end of the CListenerManager vector.
 //-----------------------------------------------------------------------------
 void CListenerManager::RegisterListener(PyObject* pCallable)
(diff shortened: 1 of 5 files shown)


src/core/modules/listeners/listeners_manager.h
```diff
@@ -33,6 +33,11 @@
 #include "utilities/wrap_macros.h"
 #include "utlvector.h"
 
+// This is required for accessing m_nFlags without patching convar.h
+#define private public
+#include "convar.h"
+#undef private
```
jordanbriere, line 39: I don't think you need that private/public hack here since you don't access any private symbols which should not produce an error when the generated object file is processed during the compilation. That said, it could probably be best to have an `utilities/convars.h` that does it and include from there instead (similar to how we have an `utilities/baseentity.h` that ensure everything is included in order, etc.) so that it is consistent across the board.

src/core/modules/listeners/listeners_manager.cpp
```diff
@@ -149,7 +161,38 @@ object CListenerManager::__getitem__(unsigned int index)
...
+	}
+}
+
+
+//-----------------------------------------------------------------------------
+// Convar changed callback.
+//-----------------------------------------------------------------------------
+void ConVarChangedCallback(IConVar* var, const char* pOldValue, float flOldValue)
+{
+	CConVarChangedListenerManager* on_convar_changed_listener_manager = GetOnConVarChangedListenerManager();
```
jordanbriere, line 176: This could be made static so that it is only retrieved once.

jordanbriere (commented):
Looks good to me. The less SP does when idled the better imo. I don't necessarily see an issue with:

> As a backwards compatibility break, when `CallGlobalChangeCallbacks` is blocked, the listener will no longer be emitted, but I don't think this will have any impact.

Since the same could be said for almost any listeners if the functions that trigger them are blocked by a hook that never forward to ours and I don't think that is something we should care about in general. If that was really causing issues, we could always lazily register our hook anyway but I don't see the benefit to that.

Question is, does it works on all games?

src/core/modules/listeners/listeners_manager.cpp
CookStar: Done!

src/core/modules/listeners/listeners_manager.h
CookStar: > I don't think you need that private/public hack here since you don't access any private symbols which should not produce an error when the generated object file is processed during the compilation.

You're right, changed it.

> That said, it could probably be best to have an `utilities/convars.h` that does it and include from there instead (similar to how we have an `utilities/baseentity.h` that ensure everything is included in order, etc.) so that it is consistent across the board.

That's a good idea. I will test it in another branch and PR it.

CookStar:
> Question is, does it works on all games?

Since InstallGlobalChangeCallback/RemoveGlobalChangeCallback are old functions (header public/icvar.h has been checked), I assume they will work, but what kind of testing do you perform when you add new functions that depend on Source SDK?

jordanbriere:
> 
> 
> > Question is, does it works on all games?
> 
> Since InstallGlobalChangeCallback/RemoveGlobalChangeCallback are old functions (header public/icvar.h has been checked), I assume they will work, but what kind of testing do you perform when you add new functions that depend on Source SDK?

AM does a pretty good job at maintaining the SDKs so I mainly only ensure that they are declared on all branches, then usually only test on CS:S/CS:GO that the features work, to finally deal with potential issues as they are reported for the others.

PR #416, opened 2021-09-23
Update outdated CS:GO Offset

Frag1337:
Found another outdated CS:GO Offset. (Has to be from a more previous update)

Credits: https://github.com/alliedmodders/sourcemod/blob/master/gamedata/sdkhooks.games/engine.csgo.txt


CookStar:
That was changed in Operation Riptide.
https://github.com/alliedmodders/sourcemod/pull/1583/files#diff-9ce230c06a7c31193cbc7efc509db4edcfb945a58c3f0b03ef361f93db0913e1R154

You missed some offsets. :)

PR #417, opened 2021-09-23
Updated CS:GO data(Operation Riptide).

CookStar:
Tested on Linux.

`is_in_field_of_view` and `blind` were checked with IDA for both Linux and Windows.

### DIFF
--- a/addons/source-python/data/source-python/entities/csgo/CBaseCombatCharacter.ini
+++ b/addons/source-python/data/source-python/entities/csgo/CBaseCombatCharacter.ini
@@ -5,8 +5,8 @@ srv_check = False
     
     # _ZNK20CBaseCombatCharacter15IsInFieldOfViewERK6Vector
     [[is_in_field_of_view]]
-        offset_linux = 274
-        offset_windows = 273
+        offset_linux = 276
+        offset_windows = 275
         arguments = POINTER
         return_type = BOOL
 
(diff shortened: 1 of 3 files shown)


Frag1337:
Good Job!

Respect for finding other offsets, e.g. increaments which weren't mentioned in sourcemod's commit.

I tried my best to just search up the changed ones from that commit, in case I couldnt find the name, I supposed it didnt existed. Sorry for the extra effort.

CookStar:
No problem!

GitHub' diff can be hard to read, sometimes it looks like it hasn't been updated, but actually it has. (e.g. https://github.com/alliedmodders/sourcemod/pull/1583/files#diff-9ce230c06a7c31193cbc7efc509db4edcfb945a58c3f0b03ef361f93db0913e1R154)

One tip on how to find changed offsets is that if a value has been updated, any offset with a larger value has a high probability of being changed. In many cases, all you need to do is run a simple test and you're good to go.

Frag1337:
Thank you for that information. I will keep that in mind for the next update ;)

But how did you manage find increament_deaths in that commit, I can't find anything related watching that link through a browser..

CookStar:
It is not in the commit, and there is no need to look for it in the commit.

> One tip on how to find changed offsets is that if a value has been updated, any offset with a larger value has a high probability of being changed. In many cases, all you need to do is run a simple test and you're good to go.

The offset we are looking for here is the offset of virtual method table/virtual function table. The reason why the offset changes is because a function is added or removed somewhere in the vtable. As an example, drop_weapon has changed this time (from 288 to 289) because a function was added below 288 in the vtable. This means that all offsets for functions above 288 most likely changed.

PR #435, opened 2021-11-11
Fixed an animation error when setting a model to Player.

CookStar:
I'm not sure if this only applies to CS:GO, but certain models seem to have problems with animation when the model is attached to player by model_index. It might be related to CCSGOPlayerAnimState::Reset(void) or something like that.

If the model is set by entity.model_index, the problem remains, but I can't decide what to do. Is it better to leave it as it is, as entity.model_name can't set the model?

Offsets other than Garry's Mod were borrowed from SourceMod, thanks to AlliedModders/SourceMod.

### DIFF
--- a/addons/source-python/packages/source-python/entities/_base.py
+++ b/addons/source-python/packages/source-python/entities/_base.py
@@ -425,16 +425,21 @@ class Entity(BaseEntity, Pointer, metaclass=_EntityCaching):
 
         return Model(model_name)
 
+    @wrap_entity_mem_func
     def set_model(self, model):
         """Set the entity's model to the given model.
 
-        :param Model model:
-            The model to set.
+        :param str/Model model:
+            The model path or model to set.
         """
-        self._set_model(model.path)
+        if isinstance(model, Model):
+            model = model.path
+
+        return [model]
 
     model = property(
-        get_model, set_model,
+        get_model,
+        lambda self, model: self.set_model(model),
         doc="""Property to get/set the entity's model.
 
         .. seealso:: :meth:`get_model` and :meth:`set_model`""")
--- a/addons/source-python/packages/source-python/entities/_base.py
+++ b/addons/source-python/packages/source-python/entities/_base.py
@@ -431,8 +431,7 @@ class Entity(BaseEntity, Pointer, metaclass=_EntityCaching):
         :param Model model:
             The model to set.
         """
-        self.model_index = model.index
-        self.model_name = model.path
+        self._set_model(model.path)
 
     model = property(
         get_model, set_model,
--- a/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
@@ -1,5 +1,11 @@
 [virtual_function]
 
+    # _ZN11CBaseEntity8SetModelEPKc
+    [[_set_model]]
+        offset_linux = 27
+        offset_windows = 26
+        arguments = STRING
+
     # _ZN11CBaseEntity9SetParentEPS_i
     [[set_parent]]
         offset_linux = 38
--- a/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
@@ -1,7 +1,7 @@
 [virtual_function]
 
     # _ZN11CBaseEntity8SetModelEPKc
-    [[_set_model]]
+    [[set_model]]
         offset_linux = 27
         offset_windows = 26
         arguments = STRING
--- a/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
@@ -4,7 +4,7 @@ srv_check = False
 [virtual_function]
 
     # _ZN11CBaseEntity8SetModelEPKc
-    [[_set_model]]
+    [[set_model]]
         offset_linux = 28
         offset_windows = 27
         arguments = STRING
--- a/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
@@ -3,6 +3,12 @@ srv_check = False
 
 [virtual_function]
 
+    # _ZN11CBaseEntity8SetModelEPKc
+    [[_set_model]]
+        offset_linux = 28
+        offset_windows = 27
+        arguments = STRING
+
     # _ZN11CBaseEntity9SetParentEPS_i
     [[set_parent]]
         offset_linux = 40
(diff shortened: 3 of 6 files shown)


addons/source-python/packages/source-python/entities/_base.py
```diff
@@ -431,8 +431,7 @@ def set_model(self, model):
         :param Model model:
             The model to set.
         """
-        self.model_index = model.index
-        self.model_name = model.path
+        self._set_model(model.path)
```
jordanbriere, line 434: Instead of having it private and forwarding the calls, I think we should simply define `_PrecacheBase.__str__`/`__int__` to return the path and index respectively so that the cast is done through extraction rather than extra layer.

CookStar:
```python
def __str__(self):
    return self._path
```

You're talking about something like this, right? But I don't think this solves the problem of private and forwarding calls. As the property `model` exists, we have to forward the call, no?

Also, it doesn't seem to be able to extract it correctly.

```
[SP] Caught an Exception:
  File "../addons/source-python/packages/source-python/events/listener.py", line 92, in fire_game_event
    callback(game_event)
  File "../addons/source-python/plugins/test/test.py", line 3227, in on_player_spawn
    player.model = player_model
  File "../addons/source-python/packages/source-python/entities/_base.py", line 291, in __setattr__
    setter(self, value)
  File "../addons/source-python/packages/source-python/entities/_base.py", line 434, in set_model
    self._set_model(model)
  File "../addons/source-python/packages/source-python/memory/helpers.py", line 335, in __call__
    return super().__call__(self._this, *args)

TypeError: No registered converter was able to extract a C++ pointer to type char from this Python object of type Model
```

jordanbriere: > You're talking about something like this, right? But I don't think this solves the problem of private and forwarding calls. As the property `model` exists, we have to forward the call, no?

Not if it is dynamically resolved from `self` (e.g. like we did [here](https://github.com/Source-Python-Dev-Team/Source.Python/pull/366#issuecomment-740217720)).



> Also, it doesn't seem to be able to extract it correctly.

It doesn't because `arg` is a raw object, but should work if casted as `boost::python::str(arg)` into [CFunction::Call](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/memory/memory_function.cpp#L303).

CookStar: That will break the compatibility of player.set_model(Model).

```
[SP] Caught an Exception:
  File "../addons/source-python/packages/source-python/events/listener.py", line 92, in fire_game_event
    callback(game_event)
  File "../addons/source-python/plugins/test/test.py", line 3232, in on_player_spawn
    player.set_model(player_model)
  File "../addons/source-python/packages/source-python/memory/helpers.py", line 335, in __call__
    return super().__call__(self._this, *args)

TypeError: No registered converter was able to extract a C++ pointer to type char from this Python object of type Model
```
Am I missing something?

jordanbriere: > That will break the compatibility of player.set_model(Model).

If `boost::python::str` is extracted as `char *`, it will fallback to resolve through `__str__` if `arg` is not already a string. So, if you cast `arg` as such before extracting, and define `Model.__str__`, `Entity.set_model(Model)` will translate to `Entity.set_model(Model.__str__)`. Another idea, though, would be to use an `EntityMemFuncWrapper` to extract the path from the passed `Model`. The rational of not having it private is that `Entity.set_model` resolves to a `MemberFunction` so that all `Function`'s functionalities are available (e.g. hooking, etc.).

CookStar: > If boost::python::str is extracted as char *, it will fallback to resolve through __str__ if arg is not already a string. So, if you cast arg as such before extracting, and define Model.__str__, Entity.set_model(Model) will translate to Entity.set_model(Model.__str__).

`player.set_model(model.__str__())` works, but `player.set_model(model)` does not.

> Another idea, though, would be to use an EntityMemFuncWrapper to extract the path from the passed Model. The rational of not having it private is that Entity.set_model resolves to a MemberFunction so that all Function's functionalities are available (e.g. hooking, etc.).

EntityMemFuncWrapper is a good idea.　Although, this could be done with model.path.

jordanbriere: > but `player.set_model(model)` does not.

It should if you change: https://github.com/Source-Python-Dev-Team/Source.Python/blob/d5da6e6f018aa2091e1303023d4d4b5d1b0a213d/src/core/modules/memory/memory_function.cpp#L303

To:

```cpp
			case DATA_TYPE_STRING:		dcArgPointer(g_pCallVM, (unsigned long) (void *) extract<char *>(str(arg))); break;
```

It also enhance `Function.__call__` by allowing any objects that can be casted as string to be passed as is.

CookStar: > It also enhance `Function.__call__` by allowing any objects that can be casted as string to be passed as is.

Ok, this is a very bad idea. At first I thought it was good, but as `str` converts all objects to strings, it can easily cause a crash when an unintended object is passed to `Function.__call__`.

jordanbriere: > Ok, this is a very bad idea. At first I thought it was good, but as `str` converts all objects to string, it can easily cause a crash when an unintended object is passed to `Function.__call__`.

Unexpected results for scripts that have a poor control over their objects? Sure, maybe. Crashes? Nah, that's a bit of an exaggeration. Worst case scenario? A string representation of the object itself is being passed to the function and likely to be rejected as an invalid input. But crashes? Very unlikely. I'm sure you would have a very hard time to find a single example to back your "easily crash" assertion here. 😆 

That said, an `EntityMemFuncWrapper` addresses the private concern of `set_model` so I'm good with that. Thanks!

CookStar: The set_model is exactly the function that causes the crash. Just to be clear, don't underestimate the nonsense of the Source Engine.

Code:
```python
player.set_model(dict)
```

Output;
```
1/ - player:  UTIL_SetModel:  not precached: <class 'dict'>
1/ - player:  UTIL_SetModel:  not precached: <class 'dict'>

1/ - player:  UTIL_SetModel:  not precached: <class 'dict'>

/home/steam/run_csgo.sh: line 16: 15772 Segmentation fault
```

jordanbriere: Hmm. That's quite stupid to terminate the whole process in such case if you ask me. Though, this raises another concern. This branch introduces `Entity.set_model(<str>)` support, meaning `Entity.set_model('not precached')` will cause a crash, right? We should add sanitation for that, so that we either;

- Remove string assignment support.
- Raise an exception.
- Pre-cache the given string (using `engine_server.precache_model` so that a `Model` don't have to be instantiated).

I think I would prefer the last option, although I don't necessarily advocate for holding scripters' hand.

CookStar: > I think I would prefer the last option, although I don't necessarily advocate for holding scripters' hand.

Since engine_server.precache_model can also cause crashes, I think it's better to Raise an exception to make the cause and consequence more obvious.

#439

PR #441, opened 2021-11-23
Changed GetEffects/SetEffects to DatamapProperty.

CookStar:
I'm not sure why it raises excption when SendTableSharedExt::find_offset returns -1, but I changed it this way because certain entities/properties need to find the offset from the datamap.

Code(CS:GO/Linux):
```python

beam = Entity.create("env_beam")
print(beam.effects)
```
Output:
```
[SP] Caught an Exception:
  File "../addons/source-python/packages/source-python/plugins/manager.py", line 207, in load
    plugin._load()
  File "../addons/source-python/plugins/test/test.py", line 4, in <module>
    print(beam.effects)

ValueError: Unable to find property 'm_fEffects'.
```
After:
```
0
```

### DIFF
--- a/src/core/modules/entities/entities_entity.cpp
+++ b/src/core/modules/entities/entities_entity.cpp
@@ -779,13 +779,13 @@ void CBaseEntityWrapper::SetDamageFilter(const char* filter)
 int CBaseEntityWrapper::GetEffects()
 {
 	static int offset = FindDatamapPropertyOffset("m_fEffects");
-	return GetDatamapPropertyByOffset<int>(offset);
+	return GetNetworkPropertyByOffset<int>(offset);
 }
 
 void CBaseEntityWrapper::SetEffects(int effects)
 {
 	static int offset = FindDatamapPropertyOffset("m_fEffects");
-	SetDatamapPropertyByOffset<int>(offset, effects);
+	SetNetworkPropertyByOffset<int>(offset, effects);
 }
 
 
--- a/src/core/modules/entities/entities_entity.cpp
+++ b/src/core/modules/entities/entities_entity.cpp
@@ -279,9 +279,11 @@ int CBaseEntityWrapper::FindNetworkPropertyOffset(const char* name)
 
 	int offset;
 	offset = SendTableSharedExt::find_offset(pServerClass->m_pTable, name);
+	if (offset == -1)
+		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "Unable to find property '%s'.", name)
 
 	// TODO: Proxied RecvTables/Arrays
-	if (offset == -1 || offset == 0)
+	if (offset == 0)
 		offset = FindDatamapPropertyOffset(name);
 
 	return offset;
@@ -776,14 +778,14 @@ void CBaseEntityWrapper::SetDamageFilter(const char* filter)
 
 int CBaseEntityWrapper::GetEffects()
 {
-	static int offset = FindNetworkPropertyOffset("m_fEffects");
-	return GetNetworkPropertyByOffset<int>(offset);
+	static int offset = FindDatamapPropertyOffset("m_fEffects");
+	return GetDatamapPropertyByOffset<int>(offset);
 }
 
 void CBaseEntityWrapper::SetEffects(int effects)
 {
-	static int offset = FindNetworkPropertyOffset("m_fEffects");
-	SetNetworkPropertyByOffset<int>(offset, effects);
+	static int offset = FindDatamapPropertyOffset("m_fEffects");
+	SetDatamapPropertyByOffset<int>(offset, effects);
 }
 
 
--- a/src/core/modules/entities/entities_entity.cpp
+++ b/src/core/modules/entities/entities_entity.cpp
@@ -279,11 +279,9 @@ int CBaseEntityWrapper::FindNetworkPropertyOffset(const char* name)
 
 	int offset;
 	offset = SendTableSharedExt::find_offset(pServerClass->m_pTable, name);
-	if (offset == -1)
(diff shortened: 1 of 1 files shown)


jordanbriere:
I don't think this is an appropriate fix, because it breaks the semantic of the function's name. We should rather just change [BaseEntity.effects' getter and setter](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/entities/entities_entity.cpp#L779-L789) to resolve from the `datamap` instead, like we did to fix similar issues for `Weapon.clip` here:

https://github.com/Source-Python-Dev-Team/Source.Python/blob/f2ed4c22ab986a4bcc7a14f072f24c17ebe069d4/src/core/modules/weapons/weapons_entity.cpp#L128-L151

CookStar:
> I don't think this is an appropriate fix, because it breaks the semantic of the function's name. We should rather just change [BaseEntity.effects' getter and setter](https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/entities/entities_entity.cpp#L779-L789) to resolve from the `datamap` instead, like we did to fix similar issues for `Weapon.clip` here:

I don't mind the change itself, but why are there separate properties/network_property/datamap_property in the first place? I understand that there is a network_property/datamap_property for those who want to access the data faster, but if the offset is not found in the network(SendTable), why not look for it in the datamap, and if the offset is not found in the datamap, why not look for it in the network (SendTable)?

For example, [GetProperty](https://github.com/Source-Python-Dev-Team/Source.Python/blob/fd63cc1e2b9b371a695bf0bc1d0687d22d0f8d79/src/core/modules/entities/entities_entity.h#L214-L230) has seen a significant performance improvement with the previous implementation.

Code:
```python
import time

beam_entity = Entity.create('env_beam')

s = time.perf_counter()
for i in range(100000):
    beam_entity.get_property_int("m_fEffects")
e = time.perf_counter()
print(e-s)
```
Before: 0.7490352485328913
After: 0.12116991728544235

jordanbriere:
> For example, [GetProperty](https://github.com/Source-Python-Dev-Team/Source.Python/blob/fd63cc1e2b9b371a695bf0bc1d0687d22d0f8d79/src/core/modules/entities/entities_entity.h#L214-L230) has seen a significant performance improvement with the previous implementation.

If the performance of `BaseEntity.<get/set>_property_<type>` is a concern, then I would rather improve them directly (e.g. by not catching a Python exception and do the offset resolution and retrieval/assignment manually, etc.). The problem I had with the original fix you proposed, was that it was rendering `FindNetworkPropertyOffset`'s name incorrect and misleading because there was no longer any guarantee the offset returned was for a networked property. Perhaps we could also add generic `FindPropertyOffset` variants that does the guesstimation, but then we wouldn't know if we have to update the edict state when assigning that offset so I'm unsure how useful it would ends up to be.

CookStar:
> If the performance of BaseEntity.<get/set>_property_<type> is a concern, then I would rather improve them directly (e.g. by not catching a Python exception and do the offset resolution and retrieval/assignment manually, etc.).

That is my plan.

> The problem I had with the original fix you proposed, was that it was rendering FindNetworkPropertyOffset's name incorrect and misleading because there was no longer any guarantee the offset returned was for a networked property.

This is the part I don't understand the most. We moved m_fEffects to datamap this time, but doesn't that cause problems with the edict state? Until now, accesses such as Player.effects were handled in NetworkProperty, but is it safe to move them to DatamapProperty?

jordanbriere:
>This is the part I don't understand the most. We moved m_fEffects to datamap this time, but doesn't that cause problems with the edict state? Until now, accesses such as Player.effects were handled in NetworkProperty, but is it safe to move them to DatamapProperty?

No, it doesn't cause problem because the retrieved offset is assigned as networked. In worst case, we flag the edict as changed and the engine complain there was no change through a warning but it shouldn't cause more issue than that.

> Is it safe to not do StateChanged() even though SendTableSharedExt::find_offset returns the correct offset?

If you don't flag the edict as changed, the new value you assigned may not be networked to clients if the server assume it didn't change which may cause predicting issues, etc.

jordanbriere:
Well, I just looked at your latest commit and yes, this will cause issue. Only the offset resolution should be handled through `datamap`, not the retrieval assignment (like it is done for `Weapon.clip` in the link I posted above).

CookStar:
I completely understand what you are saying. My bad.

> EDIT: Well, I just looked at your latest commit and yes, this will cause issue. Only the offset resolution should be handled through datamap, not the retrieval/assignment (like it is done for Weapon.clip in the link I posted above).

Fixed an embarrassing mistake.

CookStar:
I found one problem, if FindNetworkPropertyOffset can't find the offset from datamap, the behavior of get/set_network_property changes based on the entity.

If you know exactly what entity you're manipulating, it might be okay, but if it's a function argument or a return value, isn't that a problem?

jordanbriere:
> FindNetworkPropertyOffset can't find the offset from datamap

`FindNetworkPropertyOffset` doesn't lookup the datamap, so I'm unsure what issue you are trying to refer to?

CookStar:
It means that it causes these two behaviors.

pass:
```python
player = Player(index)
player.get_network_property_int("m_fEffects")
```

raises exception:
```python
entity = Entity.create('env_beam')
entity.get_network_property_int("m_fEffects")
```

jordanbriere:
Which is correct. Classes are not guaranteed to have the same properties, as subclasses may flag specific properties as excluded, or simply not declare their base class at all if it doesn't need to have any of its properties networked. Environmental entities such as `env_beam` are usually handled through TE on the clients so I would not be surprised if that entity just don't network `CBaseEntity` whatsoever. You can probably verify that through its `server_class.table` attribute.

PR #443, opened 2021-12-03
Adds entity collisions functionalities.

jordanbriere:
```python
from listeners import OnPlayerCollision

@OnPlayerCollision
def on_player_collision(player, entity):
    """Called when a player is about to collide with an entity."""
    # Disable teammates collisions
    if not entity.is_player():
        return
    return player.team_index != entity.team_index
```

```python
from listeners import OnEntityCollision

@OnEntityCollision
def on_entity_collision(entity, other):
    """Called when a non-player entity is about to collide with another entity."""
    # Disable weapons/projectiles collisions with everything except players
    return not (entity.is_weapon() and not other.is_player())
```

```python
from entities.collisions import CollisionHash
from events import Event

h = CollisionHash()

@Event('player_say')
def player_say(game_event):
    player = Player.from_userid(game_event['userid'])
    entity = player.view_entity
    if entity is None:
        return
    # Toggle collisions with aimed entity
    if h.has_pair(player, entity):
        h.remove_pair(player, entity)
    else:
        h.add_pair(player, entity)
```

```python
from entities.collisions import CollisionSet
from events import Event

s = CollisionSet()

@Event('player_say')
def player_say(game_event):
    player = Player.from_userid(game_event['userid'])
    # Toggle collisions with everything
    if player in s:
        s.remove(player)
    else:
        s.add(player)
```

```python
from entities.collisions import CollisionMap
from events import Event

m = CollisionMap()

@Event('player_say')
def player_say(game_event):
    player = Player.from_userid(game_event['userid'])
    entity = player.view_entity
    if entity is None:
        return
    # Toggle one-way collisions with aimed entity
    s = m[player]
    if entity in s:
        s.remove(entity)
    else:
        s.add(entity)
```

```python
from entities.collisions import CollisionMap
from entities.collisions import CollisionMode
from events import Event

# Setting the collision mode to ALLOW rather than PREVENT will negate the rules.
m = CollisionMap(CollisionMode.ALLOW)

@Event('player_say')
def player_say(game_event):
    player = Player.from_userid(game_event['userid'])
    entity = player.view_entity
    if entity is None:
        return
    s = m[player]
    if entity in s:
        s.remove(entity.index)
    else:
        # Player will now collide with nothing except this entity
        s.add(entity)
```

```python
from engines.trace import ContentFlags
from entities.collisions import CollisionHook

@CollisionHook
def collision_hook(entity, other, trace_filter, mask):
    """Called whenever two entities are tested for colliding contents."""
    # Prevent hostages from being killed by bullets
    if not mask & ContentFlags.HITBOX:
        return
    return other.classname != 'hostage_entity'
```

### DIFF
--- a/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
@@ -51,8 +51,8 @@
 
     # _ZNK11CBasePlayer25PhysicsSolidMaskForEntityEv
     [[get_solid_mask]]
-        offset_linux = 181
-        offset_windows = 180
+        offset_linux = 180
+        offset_windows = 179
         return_type = UINT
 
 
--- a/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/bms/CBaseEntity.ini
@@ -49,6 +49,12 @@
         offset_windows = 109
         arguments = POINTER
 
+    # _ZNK11CBasePlayer25PhysicsSolidMaskForEntityEv
+    [[get_solid_mask]]
+        offset_linux = 181
+        offset_windows = 180
+        return_type = UINT
+
 
 [input]
 
--- a/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
+++ b/addons/source-python/data/source-python/entities/csgo/CBaseEntity.ini
@@ -51,3 +51,10 @@ srv_check = False
         offset_linux = 105
         offset_windows = 104
         arguments = POINTER
+
+    # _ZNK11CBasePlayer25PhysicsSolidMaskForEntityEv
+    # https://gist.github.com/jordanbriere/679e6d09ebaa9830979b6f60e61d4b9c
+    [[get_solid_mask]]
+        offset_linux = 169
+        offset_windows = 168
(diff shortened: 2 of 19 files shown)


jordanbriere:
> does that mean the correct OnEntityCollision will be returned for the next tick?

Yes.

jordanbriere:
Just to clarify; if `CCollisionManager::ShouldHitEntity` is called, then that means the game allows the collision. If a `CollisionHook` (that is always called, regardless of the cache) prevent the collision then the cache for the current tick is updated. If the original `m_pExtraShouldHitCheckFunction` function prevents the collision, regardless of the cache, the cache is not updated and this decision takes precedence over any `CollisionHash`/`OnEntityCollision`/`OnPlayerCollision` callbacks. Basically, the cache is only used to limit `CollisionHash` lookups, as well as python calls to the listeners, and should not affect anything else beside being a major improvement.

CookStar:
I'll leave this for the record and warning, CollisionHash will cause a crash if you add more than 256 entity combinations.

**You should STOP using EntityIter or similar methods to add entities to CollisionHash.**

See the comments for more details.
https://github.com/Source-Python-Dev-Team/Source.Python/pull/444#issuecomment-989106572
https://github.com/Source-Python-Dev-Team/Source.Python/pull/444#issuecomment-989234996

PR #444, opened 2021-12-05
Enhanced CollisionHash/CollisionFilter.

CookStar:
This makes it possible to manipulate collisions in both directions.
If you have a better function name, please suggest it.

### DIFF
--- a/addons/source-python/packages/source-python/entities/collisions.py
+++ b/addons/source-python/packages/source-python/entities/collisions.py
@@ -19,30 +19,30 @@ from engines.trace import ContentMasks
 # =============================================================================
 # Source.Python Imports
 #   Entities
-from _entities._collisions import BaseCollisionHash
 from _entities._collisions import CollisionHash
 from _entities._collisions import CollisionManager
+from _entities._collisions import CollisionTable
 from _entities._collisions import collision_manager
 
 
 # =============================================================================
 # >> ALL DECLARATION
 # =============================================================================
-__all__ = [
-    'BaseCollisionHash',
-    'CollisionHash',
-    'CollisionHook',
-    'CollisionManager',
-    'collision_manager',
-]
+__all__ = ('SOLID_MASKS',
+           'CollisionHash',
+           'CollisionHook',
+           'CollisionManager',
+           'CollisionTable',
+           'collision_manager',
+           )
 
 
 # =============================================================================
 # >> INITIALIZATION
 # =============================================================================
-# Inject AutoUnload into BaseCollisionHash's hierarchy.
-if not issubclass(BaseCollisionHash, AutoUnload):
-    BaseCollisionHash.__bases__ = (AutoUnload,) + BaseCollisionHash.__bases__
+# Inject AutoUnload into CollisionHash's hierarchy.
+if not issubclass(CollisionHash, AutoUnload):
+    CollisionHash.__bases__ = (AutoUnload,) + CollisionHash.__bases__
 
 
 # =============================================================================
--- a/addons/source-python/packages/source-python/entities/collisions.py
+++ b/addons/source-python/packages/source-python/entities/collisions.py
@@ -19,7 +19,7 @@ from engines.trace import ContentMasks
 # =============================================================================
 # Source.Python Imports
 #   Entities
-from _entities._collisions import CollisionHash
+from _entities._collisions import CollisionFilter
 from _entities._collisions import CollisionManager
 from _entities._collisions import CollisionTable
 from _entities._collisions import collision_manager
@@ -29,7 +29,7 @@ from _entities._collisions import collision_manager
 # >> ALL DECLARATION
 # =============================================================================
 __all__ = ('SOLID_MASKS',
-           'CollisionHash',
+           'CollisionFilter',
            'CollisionHook',
            'CollisionManager',
            'CollisionTable',
@@ -40,9 +40,9 @@ __all__ = ('SOLID_MASKS',
 # =============================================================================
 # >> INITIALIZATION
 # =============================================================================
-# Inject AutoUnload into CollisionHash's hierarchy.
-if not issubclass(CollisionHash, AutoUnload):
-    CollisionHash.__bases__ = (AutoUnload,) + CollisionHash.__bases__
+# Inject AutoUnload into CollisionFilter's hierarchy.
+if not issubclass(CollisionFilter, AutoUnload):
+    CollisionFilter.__bases__ = (AutoUnload,) + CollisionFilter.__bases__
 
 
 # =============================================================================
--- a/src/core/modules/entities/entities_collisions.cpp
+++ b/src/core/modules/entities/entities_collisions.cpp
@@ -40,7 +40,6 @@
 // Externals.
 //-----------------------------------------------------------------------------
 extern IEngineTrace *enginetrace;
-extern IPhysics *physics;
 extern IServerTools *servertools;
 extern CGlobalVars *gpGlobals;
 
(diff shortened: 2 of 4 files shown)


jordanbriere (changes requested):
Thanks, but please leave the current implementation as is and inherit from `ICollisionHash` if you want to implement a different behaviour (just like we discussed on the forums). I prefer multiple classes that are simple and behave distinctly than a large one that tries to be them all. Moreover, it doesn't seem to currently work properly (e.g. `has_pair` always seems to return `True` for some reason).

**EDIT:** Just to add more details;

- The engine provides tools used for that very purpose; no need to reinvent the wheel.
- If we ever want to support physics collisions; it is already capable of.
- Your proposal is not even a "hash" anymore.

So, I would rather have the following 3 classes with explicit distinct behaviours:

```python
h = CollisionHash()

# entity and other never collide with each other
h.add_pair(entity, other)



m = CollisionMap()

# entity never collide with other, but other collides with entity
m[entity][other] = True



s = CollisionSet()

# entity never collide with anything
s.add(entity)
```

I would also prefer them to be kept simple (e.g. no need for the from south, from north, to east, etc. methods; they are just duplicates of the indexing operator and are redundant).

Though, the current interface doesn't really make such implementation easy, but it could be renamed/simplified to something like the following:

```cpp
class ICollisionRules
{
public:
	ICollisionRules();
	virtual ~ICollisionRules();

	virtual void OnEntityDeleted(void *pEntity) = 0;
	virtual bool ShouldHitEntity(void *pEntity, void *pOther) = 0;

	virtual void UnloadInstance();
};
```

That said, if you don't want to implement it that way, that's fine, no problem. I may just do it myself if there is ever demand for such features but for now, the current implementation fits the needs of the original feature request so I don't plan on focusing on any of that but yeah, this is the overall concept I would prefer and rather go with.

CookStar:
> * If we ever want to support physics collisions; it is already capable of.

I chose bitvec because memory usage was a concern. 
[This](https://github.com/Source-Python-Dev-Team/Source.Python/pull/443/commits/dc08465d16c8af2548d8b861459dcb574fece230) is fine when there are only a few entities to filter, but when it comes to passing all entities in the server, `boost::unordered_map<void *, boost::unordered_set<void *>>` will quickly become a memory hungry mess. However, if we want to include physics collisions, our options are limited indeed.

jordanbriere:
> CCollisionHash/ObjectPairHash has a serious flaw, which is that it cannot add a sufficient number of pairs. In order to properly process entity pairs, we need to be able to store [2,096,128](https://www.calculatorsoup.com/calculators/discretemathematics/combinations.php?n=2048&r=2&action=solve) different pairs, but we are not even 10% of that number.

Yes, I'm aware of this limitation. And just like everything else; this is not an issue until it becomes one. I doubt anyone will ever need that much space, and if they do, I will address it then.

> Features such as CollisionSet do not match that.

`CollisionSet` is more adapted for this, because it affects all current and future entities. It is a global collision state tied to the entity itself, rather than having to constantly loop through all others. And such design is redundant anyways; just remove the object itself if you want to disable them all. No need for duplicated features.

> We should develop APIs for a wide range of use cases

That's the opposite. It's always better to start with a simple and minimalist API that covers the current demand, and extend it with future requests. Trying to guess and cover unlimited of use cases is a good way to waste time developing features nobody will ever use and make the whole thing unnecessary complicated, Don't be a feature creep.

In any cases, my current branch works exactly how I wanted it to, and I won't debate about it. If there are any issues being reported, or extra features being requested, I will address them in due time. But for now, I'm very satisfied with its simplicity, usability, and flexibility.

Thanks.

CookStar:
> Yes, I'm aware of this limitation. And just like everything else; this is not an issue until it becomes one. I doubt anyone will ever need that much space, and if they do, I will address it then.

How can you tolerate a crash at less than 300 entities? It is not hard to imagine that [this](https://github.com/KoreanDude/csgo-trikz/blob/ce63d7c03d4c88a72fb8a28a295efbaa767b8a32/addons/sourcemod/scripting/trikz_solid.sp#L1550-L1618) could easily exceed 300 entities. This is completely unacceptable.

> `CollisionSet` is more adapted for this, because it affects all current and future entities. It is a global collision state tied to the entity itself, rather than having to constantly loop through all others. And such design is redundant anyways; just remove the object itself if you want to disable them all. No need for duplicated features.

How exactly do you remove a specific player or projectile from a `CollisionSet` for other player?
I would love to know how to [achieve this](https://github.com/KoreanDude/csgo-trikz/blob/ce63d7c03d4c88a72fb8a28a295efbaa767b8a32/addons/sourcemod/scripting/trikz_solid.sp#L1550-L1618) other than iterating over all entities.

> That's the opposite. It's always better to start with a simple and minimalist API that covers the current demand, and extend it with future requests. Trying to guess and cover unlimited of use cases is a good way to waste time developing features nobody will ever use and make the whole thing unnecessary complicated, Don't be a feature creep.

It is exactly you who is making it a feature creep, and it is obvious that CollisionHash/CollisionSet is unnecessary in the first place.

> In any cases, my current branch works exactly how I wanted it to, and I won't debate about it. If there are any issues being reported, or extra features being requested, I will address them in due time. But for now, I'm very satisfied with its simplicity, usability, and flexibility.
> 
> Thanks.

One last thing, is Source-Python-Dev-Team/Source.Python your personal project? If you don't accept my opinion or refuse to discuss it, that's your right, but it's not fair to unilaterally close the discussion on an issue that affects Source.Python as a whole.

jordanbriere:
> How can you tolerate a crash at less than 300 entities? It is not hard to imagine that [this](https://github.com/KoreanDude/csgo-trikz/blob/ce63d7c03d4c88a72fb8a28a295efbaa767b8a32/addons/sourcemod/scripting/trikz_solid.sp#L1550-L1618) could easily exceed 300 entities. This is completely unacceptable.

Well, that's not "_300 entities_". That's "_300 entities paired with each others_"; which is quite different. In a real scenario, not one that purposely overflow it, this limit should not really be an issue because in the end, that's quite a lot of pairs to override the collisions of. Though, I will push a fix later to address it. I haven't poked around it yet but it should be very easy to implement with a set of pairs and a custom hasher (so that they match in both directions). Which should also improve `get_pairs` anyways (the stack allocation can be quite inefficient).

> **How exactly do you remove a specific player or projectile from a `CollisionSet` for other player?**
> I would love to know how to [achieve this](https://github.com/KoreanDude/csgo-trikz/blob/ce63d7c03d4c88a72fb8a28a295efbaa767b8a32/addons/sourcemod/scripting/trikz_solid.sp#L1550-L1618) other than iterating over all entities.

Entities that are contained into a `CollisionSet` have their collision disabled with everything (current entities, future entities, etc.) as long as they are still into it. This is an isolated global state that don't discriminate with what the entities are about to collide with. Which is exactly what the code you linked seems to do at first glance.

The only reason why you need to loop through all entities in your implementation, is because you keep a Boolean state for every entities for every others and that at all time. Which, in my opinion, is unnecessary. The lack of presence is enough to tell you whether you should allow their collisions or not.

> It is exactly you who is making it a feature creep, and it is obvious that CollisionHash/CollisionSet is unnecessary in the first place.

They have specific behaviours, for different needs.

- `CollisionHash`; both-ways collisions.
- `CollisionMap`; one-way collisions.
- `CollisionSet`; global collisions.

Explicit purposes allowing scripters to pick the one that fit their needs. Which to me can be described with one word; simplicity. A huge class that tries to be them all is much less straight-forward than simple ones that have very specific and targeted usages.

> One last thing, is Source-Python-Dev-Team/Source.Python your personal project? If you don't accept my opinion or refuse to discuss it, that's your right, but it's not fair to unilaterally close the discussion on an issue that affects Source.Python as a whole.

The thing is that you didn't create an issue. You made a PR to request changes to be merged into my branch. Changes that goes directly against the way I told you beforehand I wanted it to be implemented so I'm really not sure what you expected.

PR #447, opened 2022-05-29
Adds menus tutorial to wiki

srpg:
Adds at wiki a tutorial how to make menu.
Paged, Simple and List menu

### DIFF
--- a/addons/source-python/docs/source-python/source/developing/module_tutorials/menus.rst
+++ b/addons/source-python/docs/source-python/source/developing/module_tutorials/menus.rst
@@ -1,0 +1,135 @@
+menus
+======
+
+This page contains tutorials about the :mod:`menus` package.
+
+
+Creating PagedMenu
+--------------------------
+
+This is a simple example to create paged menu
+
+.. code-block:: python
+
+    import random
+
+    from commands.say import SayCommand
+
+    from menus import PagedMenu
+    from menus import PagedOption
+
+    from players.entity import Player
+
+
+    # Register menu command
+    @SayCommand(['menu', '/menu', '!menu'])
+    def say_command(command, index, teamonly):
+        # Send the menu
+        menu.send(index)
+        return False
+
+    def my_select_callback(menu, index, option):
+        '''
+        Called whenever a selection was made.
+        '''
+    
+        # Shuffle the menu : D
+        random.shuffle(menu)
+
+        # Make it sticky
+        return menu
+
+    menu = PagedMenu(
+    title='Welcome menu',
+    description='Choose an option:',
+    select_callback=my_select_callback
+    )
+
+    # Add options from 1 to 20
+    for i in range(1, 20):
+        menu.append(PagedOption(f'{i}', i))
+
+    # Register close button to send back the menu
+    @menu.register_close_callback
+    def _on_close_menu(menu, index):
+        menu.send(index)
+
+
+Creating SimpleMenu
+--------------------------
+
+This is a simple example to create simple menu
+
+.. code-block:: python
+
+    import time
+
+    from commands.say import SayCommand
+
+    from menus import SimpleMenu
+    from menus import SimpleOption
+    from menus import Text
+
+    from messages import SayText2
+
+    from players.entity import Player
+
+    # Register menu command
+    @SayCommand(['menus', '/menus', '!menus'])
+    def say_menus_command(command, index, teamonly):
+        # Send the menu
+        menu.send(index)
+        return False
+
+    def my_menu_select_callback(menu, index, option):
+        '''
+        Called whenever a selection was made.
+        '''
+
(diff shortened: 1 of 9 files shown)


addons/source-python/docs/source-python/source/developing/module_tutorials/menus.rst
```diff
@@ -0,0 +1,116 @@
...
+    def my_menu_select_callback(menu, index, option):
+        choice = option.value
+        # Create player object from index
+        player = Player(index)
+        # Player did select yes option
+        if choice == 'yes':
+            SayText2('Thank you for accepting the rules!').send(index)
+        # Player selected no option
+        else:
+            player.kick('You have to accept the rules!')
```
jordanbriere, line 90: This should rather be:

```python
    def my_menu_select_callback(menu, index, option):
        if option.value == 'yes':
            SayText2('Thank you for accepting the rules!').send(index)
        # Player selected no option
        else:
            Player(index).kick('You have to accept the rules!')
```

As there is no reason to locally assign values to your frame unless you use them multiple times. Moreover, there is no reason to instantiate a `Player` unless your condition is not met so that is not something you should do otherwise.

jordanbriere (commented):
Thanks for the PR! Here are some points I've noticed while hovering the changes.

That said, I'm not sure how I feel about the examples themselves. Menus should be globalized and re-used, and these examples currently build them on-the-fly which encourages bad practices in my opinion. Perhaps we could leverage the examples that were provided on the forums which also happens to demonstrate more concepts and features:

https://forums.sourcepython.com/viewtopic.php?f=5&t=567

addons/source-python/docs/source-python/source/developing/module_tutorials/menus.rst
srpg: I have now updated the wiki tutorial based of your comments.
Should in wiki menu tutorial include using WeaponClassIter for making buymenu?
As example looping the weapons showing for their name and their price and it would check player can afford to buy it

addons/source-python/docs/source-python/source/developing/module_tutorials/menus.rst
```diff
@@ -0,0 +1,117 @@
+Menus
```
jordanbriere, line 1: This should be all lowercase, so that the presentation is consistent with the other tutorials on the index:

Index

addons/source-python/docs/source-python/source/developing/module_tutorials/menus.rst
```diff
@@ -0,0 +1,117 @@
...
+    def say_command(command, index, teamonly):
+        # Send the menu
+        menu.send(index)
+        return False
+
+    menu = PagedMenu(title='Welcome menu', description='Choose an option:', select_callback=my_select_callback)
+
+    # Add options from 1 to 20
+    for i in range(1, 20):
+        menu.append(PagedOption(f'{i}', i))
```
jordanbriere, line 35: This should be declared after `my_select_callback` because it currently won't load and produce a `NameError`. Moreover, the first line is too long and does not conform with [PEP 8](https://peps.python.org/pep-0008/), which we try to follow and promote for this project.

addons/source-python/docs/source-python/source/developing/module_tutorials/menus.rst
```diff
@@ -0,0 +1,117 @@
...
+    # Tell the current time
+    menu.append(Text(f"Current Time: {time.strftime('%H:%M:%S')}"))
+    # Add empty line
+    menu.append(Text(' '))
+    menu.append(Text('Do you accept the rules?'))
+    menu.append(Text(' '))
+    # Add in menu options
+    menu.append(SimpleOption(1, 'Yes', 'yes'))
+    menu.append(SimpleOption(2, 'No', 'no'))
+    menu.select_callback=my_menu_select_callback
```
jordanbriere, line 86: Same as above; this should be declared after `my_menu_select_callback`. And it should also have spaces surrounding the operator:

```python
variable = value
```

jordanbriere (commented):
> I have now updated the wiki tutorial based of your comments.

Thank you! Here are some minor points.

> Should in wiki menu tutorial include using WeaponClassIter for making buymenu?
As example looping the weapons showing for their name and their price and it would check player can afford to buy it

I don't think wiki tutorials should showcase more than minimal examples. Both for simplicity and straightforwardness so that they don't bifurcate too much into off-topic modules or become too much games specific. Is there anything pertaining to menus such example would demonstrate that is not already by the current ones?

In any case, this sounds like a good idea for the [Code examples / Cookbook](https://forums.sourcepython.com/viewforum.php?f=31) section, though.

Ayuto:
Since your branch has been messed up with the game data updates, I have now copied the tutorial manually and updated it a little bit. Thank you!

Issue #475, opened 2023-05-21
Property "desired_player_class" doesn't work for "dod"


dronelektron:
I want to get "desired_player_class" using the following code:

```Python
from commands import CommandReturn
from commands.client import ClientCommand
from players.entity import Player


@ClientCommand('sp_desired_class')
def print_desired_class(command, index):
    player = Player(index)
    desired_class = player.desired_player_class

    print(f'[DEBUG] Desired player class - {desired_class}')

    return CommandReturn.BLOCK
```

But I get the error:

```
[Source.Python]
[SP] Caught an Exception:
Traceback (most recent call last):
  File "..\addons\source-python\packages\source-python\commands\auth.py", line 44, in __call__
    return self.callback(*args)
  File "..\addons\source-python\plugins\dod_desired_class\dod_desired_class.py", line 9, in print_desired_class
    desired_class = player.desired_player_class

ValueError: Unable to find property 'm_Shared.m_iDesiredPlayerClass'.
```

I guess it's because of this:

https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/src/core/modules/players/players_entity.cpp#L746

Then I've tried to add this property to this file

https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/data/source-python/entities/orangebox/dod/CDODPlayer.ini

Like this

```
[property]

    ragdoll = m_hRagdoll
    # DOD:S has another path to "desired player class"
    desired_player_class = m_Shared.dodsharedlocaldata.m_iDesiredPlayerClass

    [[eye_angle]]
        name = m_angEyeAngles[0]
        type = QAngle
```

But this property doesn't work either, looks like SP won't override it

CookStar:
I have added a fix (#477) for this problem, but if you rewrite the ini to something like dod_desired_player_class, then you should be able to use the desired_player_class(i.e. `Player.dod_desired_player_class`) without requiring the fix.

dronelektron:
Also, this file also contains `player_class` and `desired_player_class`, but do those properties work or are they also taken from "C++" code? Just wondering.

https://github.com/Source-Python-Dev-Team/Source.Python/blob/master/addons/source-python/data/source-python/entities/orangebox/tf/CTFPlayer.ini

CookStar:
The ini is a remnant before C++ native was implemented.

PR #477, opened 2023-05-22
Added support for Player.desired_player_class to DOD:S. (#475)

CookStar:
This should solve the problem in #475. Only compiled, not tested.

### DIFF
--- a/addons/source-python/data/source-python/entities/orangebox/dod/CDODPlayer.ini
+++ b/addons/source-python/data/source-python/entities/orangebox/dod/CDODPlayer.ini
@@ -1,6 +1,7 @@
 [property]
 
     ragdoll = m_hRagdoll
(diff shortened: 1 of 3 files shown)


addons/source-python/data/source-python/entities/orangebox/dod/CDODPlayer.ini
```diff
@@ -1,6 +1,7 @@
 [property]
 
     ragdoll = m_hRagdoll
+    desired_player_class = m_Shared.dodsharedlocaldata.m_iDesiredPlayerClass
```
Ayuto, line 4: This shouldn't be required as it's implemented on C++ side. The rest looks good, thank you!

CookStar: Done.

dronelektron:
And for this please: `CDODPlayer.m_Shared.dodsharedlocaldata.m_iPlayerClass`

Issue #481, opened 2023-06-28
Pointer.set_pointer(None) causes Segmentation fault.


CookStar:
Isn't this supposed to be set to 0 (NULL)?

Code:
```python
from memory import alloc

pointer = alloc(4, False)
pointer.set_pointer(None)
```
Output:
```
Thread 1 "srcds_linux" received signal SIGSEGV, Segmentation fault.
0xf0441c67 in CPointer::SetPtr(boost::python::api::object, int) ()
   from /home/steam/csgo_ds/csgo/addons/source-python/bin/core.so
=> 0xf0441c67 <_ZN8CPointer6SetPtrEN5boost6python3api6objectEi+71>:     8b 48 04        mov    ecx,DWORD PTR [eax+0x4]
```

Ayuto:
Definitely a bug, but Jordan is right regarding the design. It's probably better to explicitly use NULL if you want to use a NULL pointer. Though, it's also quite convenient to use None.

PR #482, opened 2023-06-29
Fixed ExtractAddress causing Segmentation fault if it received None object. (#481)

CookStar:
#481

### DIFF
--- a/src/core/modules/filters/filters_recipients.h
+++ b/src/core/modules/filters/filters_recipients.h
@@ -114,7 +114,7 @@ public:
 		object self = cls();
 		object add_recipient = self.attr("add_recipient");
 
-		IRecipientFilter *pFilter = (IRecipientFilter *)ExtractAddress(oPtr);
+		IRecipientFilter *pFilter = (IRecipientFilter *)ExtractAddress(oPtr, true);
 		for (int i=0; i < pFilter->GetRecipientCount(); i++) {
 			add_recipient(pFilter->GetRecipientIndex(i));
 		}
--- a/src/core/modules/memory/memory_utilities.h
+++ b/src/core/modules/memory/memory_utilities.h
@@ -65,6 +65,9 @@
 // ============================================================================
 inline unsigned long ExtractAddress(object oPtr, bool bValidate = false)
 {
+	if (oPtr.is_none())
+		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "None was passed, expected Pointer(0) or NULL.")
+
 	CPointer* pPtr;
 
 	extract<CPointer *> extractor(oPtr);
--- a/src/core/modules/memory/memory_utilities.h
+++ b/src/core/modules/memory/memory_utilities.h
@@ -65,9 +65,6 @@
 // ============================================================================
 inline unsigned long ExtractAddress(object oPtr, bool bValidate = false)
 {
-	if (oPtr.is_none())
-		BOOST_RAISE_EXCEPTION(PyExc_ValueError, "None was passed, expected Pointer(0) or NULL.")
-
 	CPointer* pPtr;
 
 	extract<CPointer *> extractor(oPtr);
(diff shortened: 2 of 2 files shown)


CookStar:
> I don't think this should raise unless `bValidate` is `true` and should simply return `0` otherwise. Main reason being for consistency with all the other functions exported (e.g. you can pass `None` as `NULL` for any argument that accept a pointer, or you will get `None` back for a call that returns `NULL`, etc.).

I thought the same thing at first, but there is one thing that clearly breaks consistency, and that is when you override the hooked function.

```python

@EntityPreHook(EntityCondition.is_player, 'give_named_item')
def pre_give_named_item(args):
    return None # None cannot override the function.
```

And those who do not clearly know the following behavior will create unnecessary confusion, unable to understand why None returns NULL elsewhere, but not with a function override.

https://github.com/Source-Python-Dev-Team/Source.Python/blob/bcea09ef14a15cefdc4a5bd01a1521337af06879/src/core/modules/memory/memory_hooks.cpp#L129-L132

CookStar:
> That's a callback, not a function. And that's by design, because `None` is automatically returned when you omit the `return` statement.

I am not talking about the callback, I'm talking about the function that is expected to return DataType.POINTER, `give_named_item` is a function that returns DataType.POINTER. You cannot return NULL (Pointer(0)) with None.

If [NULL](https://github.com/Source-Python-Dev-Team/Source.Python/blob/bcea09ef14a15cefdc4a5bd01a1521337af06879/src/core/modules/memory/memory_wrap.cpp#L1052) exists to serve as NULL, then everything should align with [NULL](https://github.com/Source-Python-Dev-Team/Source.Python/blob/bcea09ef14a15cefdc4a5bd01a1521337af06879/src/core/modules/memory/memory_wrap.cpp#L1052). If not, then why does [NULL](https://github.com/Source-Python-Dev-Team/Source.Python/blob/bcea09ef14a15cefdc4a5bd01a1521337af06879/src/core/modules/memory/memory_wrap.cpp#L1052) exist?
https://github.com/Source-Python-Dev-Team/Source.Python/blob/bcea09ef14a15cefdc4a5bd01a1521337af06879/src/core/modules/memory/memory_wrap.cpp#L1052

CookStar:
> It's by design. Calling a Python function that does not return anything, for example:
> 
> ```python
> def pre_give_named_item(args):
>     do_whatever_but_does_not_return_anything()
> ```
> 
> Automatically return `None`. How would you differentiate a callback that just want to listen with one that want to override the returned value of the original function? In such case, this is where `NULL` becomes useful because you explicitly tell the hook manager that you want to override the call with your own value.
> 
> Regardless, there is a clear distinction between a function and a callback. When `None` is passed to a function, it should be interpreted as `NULL` for consistency with every other functions, and more importantly; backward compatibility. Changing the behaviour now would break many existing codes/features.

It is not by design, both are just compromises made due to the constraints imposed by `Python` and `Boost.Python`.

The function arguments provided by `Source.Python` accept None because `Boost.Python` foolishly associated None with NULL, and that wound should not be expanded. The type of None is never checked, this means that anytime you mistakenly pass None to a function, you are in danger of crashing.

```python
make_object(Vector, None) # Crash!

Entity.create(None) # Crash!

ptr.set_string_array(None) # Crash!
```

It should be considered an exception to the function provided by Boost.Python and should not get users into the habit of setting NULL with None. NULL should be used in situations where NULL(Pointer(0)) can be used.

Ayuto:
Oops, linked the wrong PR. I actually meant #481.

For now, the fix should be sufficient. In the past we allowed None to be used instead of NULL for convenience. The main reason why NULL exists are pre hooks. As you know you can return a value unequal to None to prevent the actual function from executing. If you do so, the returned value is returned to the caller. 

If a function does not return anything in Python, None is returned implicitly. So, in the hook handler we check against None and allow the original function to be executed. But this is a problem if you really want to return a NULL pointer. That's why we added NULL, which can be used in these cases.

I do agree that explicitly passing NULL instead of None would make it more safe. Then we also need to update our hook handler to receive a tuple containing two values: a hook action and the value that should be returned to the caller. Or maybe some classes like this:
```python
def my_hook1(args):
    return HookAction.Continue() # or None

def my_hook2(args):
    return HookAction.Override(my_value)
```

PR #496, opened 2024-09-30
Removed unnecessary _BaseSound._unload_instance function.

CookStar:
This fixes the following error when unloading the plugin

Output:
```
[Source.Python]
[SP] Caught an Exception:
  File "..\addons\source-python\packages\source-python\plugins\manager.py", line 437, in _unload_auto_unload_instances
    instance._unload_instance()
  File "..\addons\source-python\packages\source-python\engines\sound.py", line 292, in _unload_instance
    self._downloads._unload_instance()
  File "..\addons\source-python\packages\source-python\stringtables\downloads.py", line 98, in _unload_instance
    _downloadables_list.remove(self)

ValueError: list.remove(x): x not in list
```

### DIFF
--- a/addons/source-python/packages/source-python/engines/precache.py
+++ b/addons/source-python/packages/source-python/engines/precache.py
@@ -113,13 +113,7 @@ class _PrecacheBase(AutoUnload):
         self._precache()
 
     def _unload_instance(self):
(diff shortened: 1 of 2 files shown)


Ayuto:
Thank you!

Issue #498, opened 2024-11-10
Repeat.total_time_remaining is simply incorrect.


CookStar:
Code:
```python
def r():
    print("time_remaining", repeat.total_time_remaining)
    print("time_elapsed", repeat.total_time_elapsed)

repeat = Repeat(r, cancel_on_level_end=True)
repeat.start(1, 10)
print("total_time_remaining", repeat.total_time_remaining)
print("total_time_elapsed", repeat.total_time_elapsed)
print("total_time", repeat.total_time)
```
Output:
```
total_time_remaining 11.0
total_time_elapsed 0.0
total_time 10
time_remaining 10.0
time_elapsed 1.0080575942993164
time_remaining 9.0
time_elapsed 2.013115167617798
time_remaining 8.0
time_elapsed 3.0181725025177
time_remaining 7.0
time_elapsed 4.023230075836182
time_remaining 6.0
time_elapsed 5.028287410736084
time_remaining 5.0
time_elapsed 6.033344984054565
time_remaining 4.0
time_elapsed 7.038402557373047
time_remaining 3.0
time_elapsed 8.043460130691528
time_remaining 2.0
time_elapsed 9.04851746559143
time_remaining None
time_elapsed 10.052574872970581
```

PR #499, opened 2024-11-10
Fixed incorrect Repeat.total_time_remaining. (#498)

CookStar:
Fix for #498.

Changed Repeat.total_time_remaining to return 0 since there is no rational reason for it to return None when there is no delay or when Repeat is finished.


CookStar:
Changed Delay to return 0 instead of None for Delay.time_remaining and Delay.time_elapsed.

Ayuto:
Thank you!

PR #501, opened 2024-11-28
Fixes an error caused when the name of BaseEntityOutput is overlapped with another DataDesc.

CookStar:
This fixes a problem with func_bomb_target not able to get BombExplode, etc.


Ayuto:
Thank you!

PR #507, opened 2025-04-17
Updated css data

srpg:
add_account(seems to still work), increment_frag_count and increment_death_count signature not updated
Updated offsets


Ayuto:
Thank you! Have you check if they are really game specific? I think css, tf2, hl2dm, dods share the same base.

srpg:
I haven't check, but i'll check them soon for other games

PR #516, opened 2025-04-21
Fixed Source.Python work other games than css, csgo, tf2

srpg:
#478 broke Source.Python for hl2dm not sure of other games


CookStar:
Thanks for finding the problem!

Issue #520, opened 2025-05-01
Gamedata update for TF2


waldotf:
May 1st update shifted some offsets around, doesn't seem to need a rebuild.

PR #521, opened 2025-05-01
Updated TF2 data

waldotf:
For #520


waldotf:
That would make sense -- do we have a nice programmatic way to check the hardcoded SP offsets against something like https://asherkin.github.io/vtable/ ? I was manually comparing and I doubt that's very correct/efficient.

PR #523, opened 2025-07-06
Fixed the eye angle property for HL2DM.

vinci6k:
It appears they've changed the property for eye angles from **m_angEyeAngles** to **hl2mpnonlocaldata.m_angEyeAngles**.


Ayuto:
Thank you!

PR #526, opened 2025-10-04
Changed "sp update" work with github for future releases

srpg:
I tested code locally and it works perfectly
When type sp update and have same version is available it returns correctly no new version available


Ayuto:
Awesome, thank you! :)

